PR #3, opened 2024-04-17
Render documentation with eleventy instead of gulp

mikemonteith:
We need a more fully-featured static site generator, so we have decided to use Eleventy This is in line with moj-frontend

Completes #2


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,20 @@
+name: Deploy Docs
```
mikemonteith, line 1: This file needs removing from PR

davidhunter08 (approved):
LGTM

davidhunter08 (approved):

PR #7, opened 2024-04-17
Docs restructuring

mikemonteith:
- Render nunjucks tags inside markdown files
- Simplify file structure
- Use markdown for content pages
- Add a 'plain' layout with title and content
- Recompile docs when our source changes (useful during development of own component nunjucks macros)


docs/_includes/layouts/plain.njk
```diff
@@ -0,0 +1,9 @@
+{% extends "./base.njk" %}
+
+{% block content %}
+  <h1 class="govuk-heading-xl">
```
davidhunter08, line 4: Should be `nhsuk-heading-xl`

davidhunter08 (changes requested):
Looks good. Just need the heading class name changing.

docs/_includes/layouts/plain.njk
mikemonteith: I've removed the class completely

davidhunter08 (approved):
LGTM

PR #48, opened 2024-04-18
Compile docs-only CSS in eleventy build

mikemonteith:
Allows us to add scss in `./docs/assets/css/app.scss` which will be included in the documentation CSS, but not in the published npm package.


davidhunter08:
Will this also work for images and JavaScript?

mikemonteith:
> Will this also work for images and JavaScript?

It should do, but I haven't tried it

davidhunter08 (approved):

Issue #5, opened 2024-04-17
Add side navigation


davidhunter08:
We need to add side navigation to the design system so that users can navigate within the different sections.

<img width="1151" alt="Screenshot 2024-04-17 at 16 31 37" src="https://github.com/nhsuk/nhsapp-frontend/assets/14331000/984b48a0-fca7-4a3d-b3f7-89f60498c7a0">

### MOJ example
<img width="282" alt="Screenshot 2024-04-17 at 16 32 09" src="https://github.com/nhsuk/nhsapp-frontend/assets/14331000/c5b8f8af-6189-4109-8923-a15c4880f88b">

https://design-patterns.service.justice.gov.uk/get-started/

**Code**
https://github.com/ministryofjustice/moj-frontend/tree/main/src/moj/components/side-navigation
https://github.com/ministryofjustice/moj-frontend/blob/main/docs/_includes/layouts/component.njk

PR #56, opened 2024-04-24
Add side navigation

davidhunter08:
https://github.com/nhsuk/nhsapp-frontend/issues/5

### DIFF
--- a/docs/_includes/layouts/partials/side-navigation.njk
+++ b/docs/_includes/layouts/partials/side-navigation.njk
@@ -7,22 +7,14 @@
         </h{{ section.heading.headingLevel | default(4) }}>
         <ul class="app-side-navigation__list">
           {%- for item in section.items %}
-            <li class="app-side-navigation__item{% if item.active %} app-side-navigation__item--active{% endif %}"{% for attribute, value in item.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
-              <a href="{{ item.href }}"{% if item.active %} aria-current="location"{% endif %}>
-                {{- item.html | safe if item.html else item.text -}}
-              </a>
-            </li>
+            {% include "./side-navigation-item.njk" %}
           {% endfor -%}
         </ul>
       {% endfor -%}
     {% else %}
       <ul class="app-side-navigation__list">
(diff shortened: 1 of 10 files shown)


davidhunter08:
@mikemonteith 

Not rendering CSS as mixin's are undefined (eg: `@include nhsuk-font($size: 16);`)

Screenshot 2024-04-24 at 10 33 30

https://github.com/nhsuk/nhsuk-frontend/blob/c4a459e9a0236746dfec18c906738961aaad7e83/packages/components/card/card.scss#L184

mikemonteith:
`@import "node_modules/nhsuk-frontend/packages/core/all.scss";` is in app.scss, does that not include mixins etc?

davidhunter08:
> @mikemonteith
> 
> Not rendering CSS as mixin's are undefined (eg: `@include nhsuk-font($size: 16);`)
> 
> Screenshot 2024-04-24 at 10 33 30
> Same problem for `@include mq($until: tablet)`.

Fixed with https://github.com/nhsuk/nhsapp-frontend/pull/57

davidhunter08:
@mikemonteith

Could you look into how to add an active class to nav items?

On the component page layout, we call in the side navigation macro from the partials folder and add this code below...

```
{% macro sideNavigation(config) %}
  {% set config = config|addActiveAttribute(page.filePathStem | url) %}

  {{ appSideNavigation(config) }}
{% endmacro %}
```

When the side nav item is active it should have the class `app-side-navigation__item--active` and look like this...

Screenshot 2024-04-24 at 13 19 32

docs/_includes/layouts/partials/side-navigation.njk
```diff
@@ -0,0 +1,30 @@
+{% macro appSideNavigation(params) %}
```
mikemonteith, line 1: Should this file live in `_includes/partials` instead of `_includes/layouts/partials`?
It's not a layout

davidhunter08: keeping it in `_includes/layouts/partials` as the side nav is used on layouts

mikemonteith (approved):

PR #65, opened 2024-04-29
Rationalise layouts

mikemonteith:
- Use `{% block main %}` for `<main>` contents
- Use `{% block body %}` for `<body>` contents
- Use `{% block container %}` for the nhsuk-width-container contents
- Create a generic sidebar layout, which is the base for the component layout

### DIFF
--- a/docs/_includes/layouts/base.njk
+++ b/docs/_includes/layouts/base.njk
@@ -14,26 +14,34 @@
     <title>{{ title }}</title>
   </head>
   <body>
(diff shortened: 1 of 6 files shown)


mikemonteith:
Untitled-2023-08-21-1002

davidhunter08:
Pages `/examples/` and `/examples/cards/` look funky but that's ok because they'll be removed when the design examples get added?

PR #89, opened 2024-06-17
Icons and update card icon examples

edwardscull:


davidhunter08 (dismissed):
Looks good. I think we just need to be able to change the pre-fixed icon used in the card link.

src/styles/icon/icons.njk
```diff
@@ -0,0 +1,173 @@
+{% macro accountFilled(class) %}
+  <svg class="{{ class }}" xmlns="http://www.w3.org/2000/svg" height="2em" width="2em" viewBox="0 0 24 24">
```
mikemonteith, line 2: Should the `nhsapp-icon` class always be present on these svg elements, without the macro user having to specify it?

docs/components/card-links.md
```diff
@@ -44,7 +44,7 @@ Only use this variation if you have found a user need for more detailed links on
 
 Icons may help users to understand the meaning of a card link. We currently only use this variation for “Switch profiles” and “Messages” links on the homepage.
 
-[embed icon example]
+{% example "cards/card-link-icon.njk" %}
 
 If you are using icons, they should have [aria labels/hidden text?] so that screen reader users also benefit from this additional prompt on the card link.
```
mikemonteith, line 49: I think in most situations we currently have in the app, the icons should just be aria hidden with no labels or hidden text for screen readers - because they don't offer any additional context to what the link text already says.
This part of the guidance will need to be reviewed

mikemonteith (commented):
Should this PR include the docs for icons themselves, or is only intended to cover the icon-inside-card-link scenario?

src/styles/icon/icons.njk
edwardscull: The vue library does not hardcode this class os was just going off of that behaviour but happy to change it, no problem. Is this class always needed @davidhunter08

davidhunter08: It is applied to every icon so it makes sense not to make the user specify it.

edwardscull: updated

mikemonteith (approved):

Issue #28, opened 2023-11-07
Timeline (illustrative list)


davidhunter08:
## What

Use this issue to discuss the [NHS App timeline component](https://design-system.nhsapp.service.nhs.uk/components/timeline/).

![Image](https://github.com/user-attachments/assets/3298b9e1-f960-4416-851c-e2c24cd0237f)

## Journeys used on

1. Request repeat prescription
2. Waiting times

## Related

[GOV.UK Design System discussion](https://github.com/alphagov/govuk-design-system-backlog/issues/105)

growe88:
Some more examples of timeline components out there.


### **Trainline**
![IMG_8390](https://github.com/nhsuk/nhs-app-design-styles-components-patterns/assets/149805186/b5a780b0-6f6a-4e72-928e-4c554735d54a)
![Group 1](https://github.com/nhsuk/nhs-app-design-styles-components-patterns/assets/149805186/9f604b07-12c4-46bb-8dd6-b75c84421395)


### **CityMapper**
![IMG_8393](https://github.com/nhsuk/nhs-app-design-styles-components-patterns/assets/149805186/c2a29df8-17e0-4373-9a79-ea033768b1b3)


### **DHL**
![IMG_2700 1](https://github.com/nhsuk/nhsapp-frontend/assets/149805186/1d3a04a1-ba3f-4f04-9795-bfa1f395cc7f)


### **Amazon**
![IMG_2700 1](https://github.com/nhsuk/nhs-app-design-styles-components-patterns/assets/149805186/aed0666f-436d-4dd6-b6a4-ce661b6e9c7f)


### **Ikea**
![IMG_4226](https://github.com/nhsuk/nhs-app-design-styles-components-patterns/assets/149805186/317b4a23-5e6a-4c9f-a148-a813897765f2)


### TFL Go
![IMG_8394](https://github.com/nhsuk/nhs-app-design-styles-components-patterns/assets/149805186/a55fa112-211e-435b-8f83-389e05478098)

Tosin-Balogun:
Prescription and medicines team use this component to inform user the stages their medicine will go through before it reaches them


![Image](https://github.com/nhsuk/nhsapp-frontend/assets/71393402/522c0660-9f03-4854-b95a-6378921f97ee)

growe88:
**Collection of assorted information pertaining to Timeline Component**

**Insights:**
[...]

PR #92, opened 2024-06-21
timeline component

edwardscull:


mikemonteith:
Not sure about the semantics of this.

"Request medication" is a heading but then it has no content below it.
Should be be looking at this being an ordered list instead?

image

mikemonteith:
We are using colour to convey the past/present/future state. Do we need some markup for screen readers to do the same?

davidhunter08:
> Not sure about the semantics of this.
> 
> "Request medication" is a heading but then it has no content below it. Should be be looking at this being an ordered list instead?
> 
> image
> Edit: I've just noticed that it _is_ an ordered list!

Good point @mikemonteith - in all the [example timelines I've seen across gov](https://github.com/nhsuk/nhsapp-frontend/issues/28), they use a heading, but I haven't seen one with a heading and no content beneath.

Maybe that example is a bad one (and a content problem?) and we need to change it? Also add guidance, eg: "it must have content beneath the heading".

davidhunter08:
> We are using colour to convey the past/present/future state. Do we need some markup for screen readers to do the same?

I'm not sure. We can include this in the accessibility testing we're planning.

src/components/timeline/_timeline.scss
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    &--past {
+      &:before {
+        border-color: $color_nhsuk-blue;
+      }
+    }
+  }
+
+  &__badge {
```
mikemonteith, line 54: Is the timeline icon a badge? We have [the badge component](https://nhsuk.github.io/nhsapp-frontend/components/badge/) already so might make sense to come up with a different name for this.

src/components/timeline/_timeline.scss
```diff
@@ -0,0 +1,97 @@
+/* ==========================================================================
```
mikemonteith, line 1: There are a few "magic numbers" in this file.
`margin-top: 4px;`, `margin-top: -6px;` etc.
Can we refactor to make these values computed from the size variables?

src/components/timeline/_timeline.scss
```diff
@@ -0,0 +1,97 @@
...
+  &__badge {
+    flex-shrink: 0;
+    height: $timeline-badge-size-mobile;
+    margin-left: - calc(($timeline-badge-size-mobile / 2) + ($timeline-border-width / 2));
+    margin-right: nhsuk-spacing(4);
+    margin-top: 4px;
+    width: $timeline-badge-size-mobile;
+    z-index: 1;
+
+    @include mq($from: tablet) {
```
mikemonteith, line 63: We could probably define a mixin to do the height, width and margin positioning for all 4 variations of the timeline icon (small & regular X mobile & tablet).

src/components/timeline/timeline.njk
```diff
@@ -0,0 +1,31 @@
...
+  {% set isPastItem = 'nhsapp-timeline__item--past' if item.isPastItem else '' %}
+  
+  <li class="nhsapp-timeline__item {{ isPastItem | trim }}">
+    {% if item.active or item.isPastItem %}
+      <svg class="nhsapp-timeline__badge" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="28" height="28" viewBox="0 0 28 28" fill="none">
+        <circle cx="14" cy="14" r="13" transform="matrix(-1.19249e-08 -1 -1 1.19249e-08 28 28)" fill="#005EB8" stroke="#005EB8" stroke-width="2"/>
+      </svg>
+    {% else %}
+      <svg class="nhsapp-timeline__badge nhsapp-timeline__badge--small" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 14 14">
+        <circle cx="7" cy="7" r="6" transform="matrix(-1.19249e-08 -1 -1 1.19249e-08 14 14)" fill="white" stroke="#AEB7BD" stroke-width="2"/>
```
mikemonteith, line 13: The `transform` attribute can be deleted from the SVGs in this file

src/components/timeline/_timeline.scss
edwardscull: updated @mikemonteith

src/components/timeline/_timeline.scss
edwardscull: As discussed offline, will need @davidhunter08 input

src/components/timeline/_timeline.scss
edwardscull: As discussed offline, will need @davidhunter08 input

davidhunter08:
Screenshot 2024-07-02 at 15 33 37

Looking at the HTML output makes me wonder if we can refactor the CSS to make it a bit cleaner. Possibly use a 'background: node-blue.svg` on the `li:before`?

What do you think @edwardscull @mikemonteith ?

edwardscull:
> Screenshot 2024-07-02 at 15 33 37
> Looking at the HTML output makes me wonder if we can refactor the CSS to make it a bit cleaner. Possibly use a 'background: node-blue.svg`on the`li:before`?
> 
> What do you think @edwardscull @mikemonteith ?

Yes that would be cleaner for sure. It's an image that probably makes sense to be a background image.

davidhunter08 (approved):

PR #99, opened 2024-07-05
Update badge guidance and examples

davidhunter08:
- update badge guidance
- add large badge example with 9+
- add messages inbox with badge example
- add bottom nav with badge image
- add truncate scss


src/all.scss
```diff
@@ -9,3 +9,6 @@
 @import "components/card/card";
 @import "components/tag/tag";
 @import "components/timeline/timeline";
+
+// utilities
+@import "./utilities/truncate";
```
edwardscull, line 14: nit: don't need the `./`

edwardscull (approved):
👍

PR #105, opened 2024-07-10
Reorder nhsuk-frontend dependency

edwardscull:
When developing locally the dev dependency gets installed

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,6 @@
     "sass": "^1.74.1"
   },
   "peerDependencies": {
-    "nhsuk-frontend": "^8.1.1"
+    "nhsuk-frontend": "^8.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -41,11 +41,12 @@
     "gulp-sass": "^5.1.0",
     "gulp-uglify": "^3.0.2",
     "markdown-it-anchor": "^9.0.1",
+    "nhsuk-frontend": "^8.1.1",
     "nunjucks": "^3.2.4",
     "prettier": "^3.2.5",
     "sass": "^1.74.1"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -41,11 +41,12 @@
     "gulp-sass": "^5.1.0",
     "gulp-uglify": "^3.0.2",
     "markdown-it-anchor": "^9.0.1",
+    "nhsuk-frontend": "^8.1.1",
     "nunjucks": "^3.2.4",
     "prettier": "^3.2.5",
     "sass": "^1.74.1"
   },
-  "dependencies": {
+  "peerDependencies": {
```
mikemonteith, line 49: Should this peerDependency have a version like "8.x" (or "^8.0.0")?
There's nothing specific in 8.1.1 that we must require our users to have.

mikemonteith (changes requested):
Do we also need to add nhsuk-frontend as a dev dependency, so that when someone checks out this repo to work on it, they have that dependency too?

package.json
edwardscull: We could make it `^8.0.0` yeh as this will just install `8.2.0`. The idea being we are betting on MINOR bumps not to affect anything for us.

package.json
```diff
@@ -41,11 +41,12 @@
     "gulp-sass": "^5.1.0",
     "gulp-uglify": "^3.0.2",
     "markdown-it-anchor": "^9.0.1",
+    "nhsuk-frontend": "^8.1.1",
```
edwardscull, line 44: @mikemonteith I have added it here in dev deps too

mikemonteith (approved):

PR #113, opened 2024-07-19
Read only card

edwardscull:
SUS Cards (non-interactive) (1)

- add `readOnly` param to template
- replace `a` tag with heading
- hide chevron icon

### DIFF
--- a/src/components/card/card.njk
+++ b/src/components/card/card.njk
@@ -21,7 +21,7 @@
         </div>
       {%- endif -%}
       {%- if params.readOnly -%}
-        <h3 class="nhsapp-card__title">{{ params.title }}</h3>
+        <p class="nhsapp-card__title">{{ params.title }}</p>
       {%- else -%}
         <a href="{{ params.href }}" class="nhsapp-card__link nhsuk-link--no-visited-state" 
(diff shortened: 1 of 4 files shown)


src/components/card/card.njk
```diff
@@ -18,9 +18,13 @@
           {{ params.aboveContent.html | safe }}
         </div>
       {%- endif -%}
-      <a href="{{ params.href }}" class="nhsapp-card__link nhsuk-link--no-visited-state">
-        {{ params.title }}
-      </a>
+      {%- if params.readOnly -%}
+        <h3 class="nhsapp-card__title">{{ params.title }}</h3>
```
edwardscull, line 22: @davidhunter08 should this heading support more than just `h3`?

davidhunter08: I did a quick syntax checker and it came up with a warning when using a `h3`:

> h3 may be used for formatting

Screenshot 2024-09-11 at 12 36 01

We [don't use headings in card links](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/249#issuecomment-2011772269) so having it as a `<p>` should be ok.

davidhunter08 (approved):

PR #120, opened 2024-07-25
Prefix macros with `nhsapp`

davidhunter08:


edwardscull (approved):
LGTM. Technically this a breaking change so we should do a major bump on the next release e.g. `v2.0.0`

PR #131, opened 2024-08-01
Symlink solution

edwardscull:


nhsapp
mikemonteith, line 1: If you add `nhsapp` at this level, then you would need to add `.` to the nunjuck include paths, so that nunjucks can resolve `nhsapp/components/...`

Maybe since this is specific to `docs`, the symlink should be inside `docs/_includes/`.

`docs/_includes/nhsapp` symlink to `src`.
Then any nunjucks include starting with `nhsapp/foo` would resolve to `src/foo`

davidhunter08 (approved):
LGTM

PR #161, opened 2024-10-09
Suggested change to introduction text

Graham-Pembrey:
I wondered if we could shorten this introduction to the community page, and get straight to talking about the 'community' aspect? The first line on the live version perhaps duplicates the homepage statement about the overall purpose of the design system. See what you think though @davidhunter08!


davidhunter08 (approved):
LGTM

PR #162, opened 2024-10-09
Small tweak to punctuation

Graham-Pembrey:
Minor suggestion to avoid using a dash. I do think the currently live version reads well, but the service manual guidance is to be cautious about using dashes. https://service-manual.nhs.uk/content/formatting-and-punctuation


davidhunter08 (approved):
LGTM

PR #169, opened 2024-10-16
Add new icons

davidhunter08:
Adding new icons:
- calendar
- clock
- download 
- email
- filter
- letter
- night
- phone call
- sort

image


davidhunter08:
`calendar` icon name should probably be `add-calendar`?

davidhunter08:
@edwardscull how do you think we should handle the colour of the icons?

All previous icons have been blue but some of the new icons (clock, night and phone call) need to be black.

Example:
image

edwardscull:
> @edwardscull how do you think we should handle the colour of the icons?
> 
> All previous icons have been blue but some of the new icons (clock, night and phone call) need to be black.
> 
> Example: image

Have we confirmed whether we need to support both blue and black?

davidhunter08:
> > @edwardscull how do you think we should handle the colour of the icons?
> > All previous icons have been blue but some of the new icons (clock, night and phone call) need to be black.
> > Example: image
> 
> Have we confirmed whether we need to support both blue and black?

Yes. I've added a modifier class `nhsapp-icon-black`.

edwardscull (approved):

PR #197, opened 2024-11-29
Update component guidance

davidhunter08:
Updating the component guidance to let teams know they can iterate components (if there is a need to) and share any changes with us in GitHub.

| Before | After |
| -------|-------|
| old | new |

### DIFF
--- a/docs/community/share-findings.md
+++ b/docs/community/share-findings.md
@@ -19,7 +19,7 @@ Useful findings to share might be things like:
 
 If the thing is already published in the NHS App design system, go to the relevant page. You'll find a link to the GitHub issue at the bottom of the page.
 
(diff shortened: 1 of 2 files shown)


michaelgallagher (approved):
approved

Graham-Pembrey (changes requested):
Looks great! One small suggestion. Could we change:

`Existing design components can be used as starting points but may need to be iterated to suit the needs of users of a particular service.`

To be in the active voice:

`You can use existing components as starting points, but you may need to iterate them to meet user needs for particular services.`

davidhunter08:
Thanks @Graham-Pembrey, I've updated the content 👍

Screenshot 2024-12-03 at 09 10 08

Graham-Pembrey (approved):
Thanks Dave. Looks good

Issue #12, opened 2023-11-02
Buttons


davidhunter08:
## What

Use buttons to help users carry out an action on a page like starting an application or saving their progress.

![Image](https://github.com/nhsuk/nhsapp-frontend/assets/14331000/f34bdcf2-b5c3-4fdb-98b2-ec020e892863)

## Related
[NHS App design system Figma file](https://www.figma.com/file/6f2CbcZ7cnpNrtKEcfQp8X/WIP-NHS-App-Design-System-Branch?type=design&node-id=128%3A4944&mode=design&t=8zxXry0rS9LIpQUk-1)
[NHS design system guidance](https://service-manual.nhs.uk/design-system/components/buttons)
[NHS design system discussion](https://github.com/nhsuk/nhsuk-service-manual-backlog/issues/7)
[GOV.UK Design System guidance](https://design-system.service.gov.uk/components/button/)
[GOV.UK Design System discussion](https://github.com/alphagov/govuk-design-system-backlog/issues/34)

## Examples on the NHS App

<details>
  <summary>Examples</summary>

Primary button
![Image](https://github.com/nhsuk/nhs-app-redesign-prototype/assets/14331000/3d44cbc5-1c93-4f58-9ad9-956bf0bebb5c)

Secondary button
![Image](https://github.com/nhsuk/nhs-app-redesign-prototype/assets/14331000/0c454a07-2f91-4c2e-b640-8d8c0394eee0)

Reverse button
![Image](https://github.com/nhsuk/nhs-app-redesign-prototype/assets/14331000/6cfb74e9-a267-4ccb-8463-0252c4208d52)

Full width (on small screens)
![Image](https://github.com/nhsuk/nhs-app-redesign-prototype/assets/14331000/dcc2cf84-6a32-4be0-9023-56db9a50a3b4)
</details>

### Design checklist

- [x] Gather evidence and consolidate designs
- [x] Create a rough proposal (Figma)
- [x] Build a working version (HTML or Native)
- [x] Draft guidance
- [x] Design sign off

### Development criteria

- [x] Make WCAG 2.2 compliant
- [x] Add HTML / Nunjucks version
- [ ] Add vue.js version
- [x] Add Figma version
- [x] Add design system guidance

Tosin-Balogun:
Prescriptions and medicines use buttons in several of their prototypes

[...]

PR #203, opened 2024-12-05
Add new secondary button

davidhunter08:
## Description

- [Update secondary button style](https://github.com/nhsuk/nhsapp-frontend/issues/12#issuecomment-2385106820)
- Update button spacing

### Button guidance

| Before | After |
| -------|-------|
| image | image |

### Button spacing

button-size-desktop

button-size-mobile

## To do
- [x] add secondary button css
- [x] update button examples
- [x] update guidance

### DIFF
--- a/docs/components/buttons.md
+++ b/docs/components/buttons.md
@@ -17,6 +17,13 @@ We use the [NHS design system button](https://service-manual.nhs.uk/design-syste
 
 ### Secondary button
 
+Our secondary button design differs from the NHS design system. The [research section](#secondary-button-1) below has more details about why.
+
+Use a secondary button either:
+
+- for secondary actions on a page
+- when an action needs less prominence – for example because it is optional
+
 {% example "buttons/button-secondary.njk" %}
 
 ### Buttons on dark backgrounds (reverse button)
@@ -35,7 +42,15 @@ Any links within a button group will automatically align with the buttons.
 
 ## Research
 
+### Full width buttons
+
 Testing on the NHS App has found that full width buttons are not a problem for users with smaller screens. But we will continue to test if:
 
 - full width buttons are problematic for users with low digital skills
 - users don’t actually see them as actionable things to click
+
+### Secondary button
+
+We designed the NHS App secondary button as an alternative to a secondary button with a grey background. This was because our research found some participants overlooked that design or believed it was a disabled button. Some participants hesitated before selecting the button. The visual hierarchy of primary and secondary buttons was also found to be unclear for people with monochromatic vision.
+
+In moderated usability testing, six out of six participants were able to use the NHS App secondary button for a variety of actions and without hesitation. Two participants had colour vision deficiency (colour blindness).
--- a/docs/examples/buttons/button-group.njk
+++ b/docs/examples/buttons/button-group.njk
@@ -16,7 +16,7 @@ vueLink: "https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?pat
 
 {{ button({
   text: "Save as draft",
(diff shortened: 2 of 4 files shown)


davidhunter08:
I've removed some code that was changing the button padding (in comparison to the NHS design system button).

The differences were:
- on mobile - top/bottom padding increases to 12px (from 8px)
- on desktop - left/right padding increases to 32px (from 16px)

button padding

davidhunter08:
The default button left/right padding (16px) looks too small because of all the extra visual space on top of the label due to the cap/x-height.

20px looks best, but 24px also looks ok and uses the standard spacing units.

32px looks too big.

button padding comparison

davidhunter08:
> The default button left/right padding (16px) looks too small because of all the extra visual space on top of the label due to the cap/x-height.
> 
> button padding comparison

I've set it to 24px for now, pending feedback

davidhunter08:
Vertical spacing comparison.

NHS design system = 8px padding (top & bottom)
NHS App design system = 12px padding (top & bottom)

button vertical padding comparison

edwardscull (approved):

davidhunter08:
## Current guidance

image

Guidance needs updating to:
- say that the secondary button is different to the NHS design system
- reference the research

Also is the current usage guidance for secondary buttons correct?

> **NHS design system usage guidance:**
> Use a secondary button on pages that have more than 1 action or when users are not noticing standard link text.

Are there instances where you would use the secondary button as the only action?

Here's the gov guidance:

> **GOV.UK design system usage guidance:**
> Use secondary buttons for secondary calls to action on a page.

davidhunter08:
[A question from Katy around secondary button usage](https://github.com/nhsuk/nhsapp-frontend/issues/12#issuecomment-2505832123):

> We currently use Secondary button + Link assembly component to give users two distinct options. Our reasoning for using secondary and link rather than primary and link is that each CTA is equally weighted and one is not the natural next step over the other, which the green button may suggest. Is this logic used in other services or is primary button + link the default here?

cc @Graham-Pembrey

Graham-Pembrey:
Draft addition to the secondary button usage guidance:

> Our secondary button design differs from the NHS design system.
> 
> Use a secondary button either:
> 
> - for secondary actions on a page
> - when an action needs less prominence – for example because it is optional

The second point would fit with cases where we might have reasons to use a secondary button as the only action on a page. It could also work in the scenario Katy described. 

Draft addition to the research section of the guidance:

> We designed the NHS App secondary button as an alternative to a secondary button with a grey background. This was because our research found some participants overlooked that design or believed it was a disabled button. In moderated usability testing, six out of six participants were able to use the NHS App secondary button for a variety of actions. Two participants had colour vision deficiency (colour blindness).

Graham-Pembrey:
Suggested changes to the draft guidance above after further discussions with the team this morning. Additions are in code blocks: 

> `We use a variation of the NHS design system secondary button. The "Research" section below has more details about why. `
> 
> Use a secondary button either:
> 
> - for secondary actions on a page
> - when an action needs less prominence – for example because it is optional

In the research section:

> We designed the NHS App secondary button as an alternative to a secondary button with a grey background. This was because our research found some participants overlooked that design or believed it was a disabled button. `Some participants hesitated before selecting the button. The visual hierarchy of primary and secondary buttons was also found to be unclear for people with monochromatic vision.`
> 
> In moderated usability testing, six out of six participants were able to use the NHS App secondary button for a variety of actions `and without hesitation`. Two participants had colour vision deficiency (colour blindness).

davidhunter08:
## Draft guidance - v2

image

Graham-Pembrey (approved):

Issue #186, opened 2024-11-12
Update code highlighter colours


davidhunter08:
Some colours used for highlighting code snippets do not meet the required contrast for accessibility.

<img width="742" alt="Screenshot 2024-11-12 at 09 34 53" src="https://github.com/user-attachments/assets/c1952cd2-86fc-4da6-bee8-416dd2a5aca8">

PR #206, opened 2024-12-11
Fix code syntax highlight colour contrast

davidhunter08:
Added `_code-colours.scss` to override prism.js colours.

Fixes https://github.com/nhsuk/nhsapp-frontend/issues/186

Screenshot 2024-12-11 at 14 31 56
Screenshot 2024-12-11 at 14 31 42

### DIFF
--- a/docs/assets/css/app.scss
+++ b/docs/assets/css/app.scss
@@ -13,6 +13,7 @@
 @import "node_modules/nhsuk-frontend/packages/components/button/button";
 
 @import "node_modules/prismjs/themes/prism";
+@import "docs/assets/css/code-colours";
 
 // Import all styles from the NHS App frontend library
 @import "src/all";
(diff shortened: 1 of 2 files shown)


docs/assets/css/app.scss
```diff
@@ -13,6 +13,7 @@
 @import "node_modules/nhsuk-frontend/packages/components/button/button";
 
 @import "node_modules/prismjs/themes/prism";
+@import "docs/assets/css/code-colours";
```
edwardscull, line 16: If this is overriding all prism colours, can line 15 just be deleted?

davidhunter08: It's not overriding all the colours and it's not just colours in the prism theme file so it's still needed.

edwardscull (approved):

Issue #229, opened 2025-01-22
Update badge colour guidance


davidhunter08:
The N&M team are changing the blue badge to a red badge on the messages inbox. Making the use a red dot consistent across the app. So we need to review and update the [badge component guidance and examples](https://design-system.nhsapp.service.nhs.uk/components/badge/).

PR #235, opened 2025-01-24
Remove small blue badge variant

davidhunter08:
Fixes https://github.com/nhsuk/nhsapp-frontend/issues/229

- changed small badge red to default in CSS
- removed if red logic from badge macro
- updated guidance and examples
- improved spacing between image examples
- added CHANGELOG entry

| Before | After |
| -------|-------|
| image | image |


Graham-Pembrey (commented):
In the "How to use" section, should we also remove the following line? 

> There are different size and colour variations of badges.

We could leave the size part in, but I think the "Large badge" and "Small badge" headings make that clear anyway.

davidhunter08:
> In the "How to use" section, should we also remove the following line?
> 
> > There are different size and colour variations of badges.
> 
> We could leave the size part in, but I think the "Large badge" and "Small badge" headings make that clear anyway.

Updated @Graham-Pembrey 👍

Graham-Pembrey (approved):

src/components/badge/small/badge-small.njk
```diff
@@ -1,6 +1,5 @@
 <span class="
     nhsapp-badge-small 
-    {{ 'nhsapp-badge-small--red' if params.color === 'red' }}
```
edwardscull, line 3: Removing this would be a breaking change. If someone is using the nunjucks macro and declaring the `color: 'red'` param, they will get a template error.

I think we can just leave it in as well as the `--red` modifier class. Then just default the badge to red like you are doing. Meaning that modifier class will style red if it is present and will still style red if it isn't. This will be a backwards compatible change.

OR

We do a major version bump? Although possibly bit overkill to bump `3.0.0` for this change.

edwardscull (changes requested):

src/components/badge/small/badge-small.njk
davidhunter08: I did test it with declaring `color: red` and it worked for me.

Screenshot 2025-01-28 at 10 06 24

Must be a quirk in nunjucks templating that it doesn't mind you declaring arguments that don't exist?

edwardscull: Ah awesome, thanks Javascript! Ok in that case 🚀

edwardscull (approved):

Issue #175, opened 2024-10-18
Bump NHS.UK frontend version (v9.2.0)


davidhunter08:
### Description

The NHS.UK frontend has been updated, including changes to:
- typography sizes
- header component (updated in v8)
- back link / breadcrumb placement
- main wrapper padding

https://github.com/nhsuk/nhsuk-frontend/releases

## Done when

- [x] [Bump NHSUK frontend version and test in NHS App frontend](https://github.com/nhsuk/nhsapp-frontend/pull/241)
- [x] [Document changes](https://github.com/nhsuk/nhsapp-frontend/pull/241)
- [x] Review changes with design / tech leads
- [ ] Merge to main and release package

PR #241, opened 2025-02-10
Bump `nhsuk-frontend` version `v9.2.0`

davidhunter08:
## Description

Bumped the version of the `nhsuk-frontend` to `v9.2.0`
- https://github.com/nhsuk/nhsapp-frontend/issues/175

Bump peer dependency `nhsuk-frontend` to `v9.0.0`

## Page layout changes

- added `{% from 'attributes.njk' import nhsukAttributes %}` to base page template (for component macros to work)
- changed `{% block breadcrumb %}{% endblock %}` to `{% block beforeContent %}{% endblock %}`
- moved `{% block beforeContent %}{% endblock %}` to within `container` before `main`
- removed `nhsuk-main-wrapper--s` class because `nhsuk-main-wrapper` padding has decreased from `40px` to `24px` on mobile

## Visual design changes

**Card links** font size on mobile changed from `18px` to `19px`

| Before | After |
| -------|-------|
| Screenshot 2025-02-10 at 15 28 54 | Screenshot 2025-02-10 at 15 28 39 |

**Section headings** font size on mobile changed from `18px` to `19px`

| Before | After |
| -------|-------|
| Screenshot 2025-02-10 at 15 54 32 | Screenshot 2025-02-10 at 15 53 57 |

## Other stuff to consider

- **global header (web)** - do we want to swap it for the new version? It changed in version 8 of the nhsuk-frontend but the app is using version 7

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nhsapp-frontend",
-  "version": "2.3.0",
+  "version": "3.0.0",
   "description": "A CSS library built on top of nhsuk-frontend providing styles for the NHS App.",
   "main": "index.js",
   "type": "module",
--- a/package.json
+++ b/package.json
@@ -43,12 +43,12 @@
     "gulp-uglify": "^3.0.2",
     "gulp-zip": "^6.0.0",
     "markdown-it-anchor": "^9.0.1",
-    "nhsuk-frontend": "^8.1.1",
+    "nhsuk-frontend": "^9.1.0",
     "nunjucks": "^3.2.4",
     "prettier": "^3.2.5",
     "sass": "^1.74.1"
   },
   "peerDependencies": {
-    "nhsuk-frontend": "^8.0.0"
+    "nhsuk-frontend": "^9.1.0"
   }
 }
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -43,12 +43,12 @@
...
     "markdown-it-anchor": "^9.0.1",
-    "nhsuk-frontend": "^8.1.1",
+    "nhsuk-frontend": "^9.1.0",
     "nunjucks": "^3.2.4",
     "prettier": "^3.2.5",
     "sass": "^1.74.1"
   },
   "peerDependencies": {
-    "nhsuk-frontend": "^8.0.0"
+    "nhsuk-frontend": "^9.1.0"
```
edwardscull, line 52: Can we make this `"9.0.0"` (given that version actually exists in the nhsuk frontend?)

edwardscull (changes requested):
I think this all looks good. We will need to do `v3.0.0` bump.

package.json
davidhunter08: Updated peer dependency to `9.0.0`

edwardscull (approved):
🚀

PR #254, opened 2025-03-04
Add sub-hub page guidance

davidhunter08:
## Draft guidance - V1

image


davidhunter08:
The sub-hub page examples don't include the 'related content' component as that doesn't exist currently in the NHS App design system. Is that ok? Should we add it to the design system?

Screenshot 2025-03-04 at 13 19 46

Graham-Pembrey (approved):

PR #276, opened 2025-03-13
Lowercase 'npm'

davidhunter08:
Lowercase 'npm' to make it consistent across guidance.

Also added `node package manager (npm)` to the first time it gets mentioned on the page, following what the service manual does.

### DIFF
--- a/docs/get-started/installing-with-npm.md
+++ b/docs/get-started/installing-with-npm.md
@@ -1,13 +1,13 @@
 ---
 layout: layouts/get-started.njk
-title: Installing NHS App frontend with NPM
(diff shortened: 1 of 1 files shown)


Graham-Pembrey (commented):
Looks good! One open question: do we need 'with npm' in the title of this page? Would 'Installing NHS App frontend' be enough (but leaving the npm mentions on the page)?

davidhunter08:
> Looks good! One open question: do we need 'with npm' in the title of this page? Would 'Installing NHS App frontend' be enough (but leaving the npm mentions on the page)?

@edwardscull any thoughts on this?

PR #277, opened 2025-03-14
Fix secondary button layout shift

davidhunter08:
Fix layout shift on secondary button :active and :focus, make :before behave same as primary button.

- https://github.com/nhsuk/nhsuk-frontend/pull/1077/commits/82173ea9b07a8205c29ea54f0e804858ee6b9d99

Increase secondary button active background colour.

- https://github.com/nhsuk/nhsuk-frontend/pull/1077/commits/c38321dfb2072ca3127fe702e49a99e06b8685aa

### DIFF
--- a/src/components/button/_index.scss
+++ b/src/components/button/_index.scss
@@ -8,7 +8,7 @@ $nhsapp-secondary-button-color: transparent;
 $nhsapp-secondary-button-border-color: $color_nhsuk-blue;
 $nhsapp-secondary-button-text-color: $color_nhsuk-blue;
 $nhsapp-secondary-button-hover-color: tint($nhsapp-secondary-button-border-color, 85%);
-$nhsapp-secondary-button-active-color: tint($nhsapp-secondary-button-border-color, 75%);
+$nhsapp-secondary-button-active-color: tint($nhsapp-secondary-button-border-color, 78%);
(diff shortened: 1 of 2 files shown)


src/components/button/_index.scss
```diff
@@ -52,6 +52,7 @@ $nhsapp-secondary-button-shadow-color: $nhsapp-secondary-button-border-color;
   }
 
   &:focus {
+    @include nhsuk-responsive-margin(5, "bottom", $adjustment: $button-shadow-size * 1.5);
```
edwardscull, line 55: What exactly was this issue here? Does the service manual also have this problem?

src/components/button/_index.scss
```diff
@@ -82,6 +84,10 @@ $nhsapp-secondary-button-shadow-color: $nhsapp-secondary-button-border-color;
       padding-bottom: $button-padding-top-bottom-mobile;
       padding-top: $button-padding-top-bottom-mobile;
     }
+
+    &::before {
+      bottom: -($nhsuk-border-width-form-element + $button-shadow-size);
```
edwardscull, line 89: Does this calculation come from somewhere specific, or is it custom?

src/components/button/_index.scss
davidhunter08: The service manual have the same problem which will be fixed when they release the new secondary button.

edwardscull (approved):

PR #313, opened 2025-04-30
Change badge component name to badges

beakking2:


beakking2:
this was a test for the 'Propose content change' guidance

PR #323, opened 2025-06-10
Update error page guidance post-review panel

Graham-Pembrey:
Some content changes to reflect the feedback from the design system review panel.

| Before | After |
| -------|-------|
| image | image |

### DIFF
--- a/docs/patterns/error-page.md
+++ b/docs/patterns/error-page.md
@@ -37,7 +37,7 @@ The page should:
 This page should not:
 
 - blame the user
-- include breadcrumbs
+- use breadcrumbs or a standard back link at the top
 - display vague terms or jargon like "500", "504", "bad request" or "we are experiencing technical difficulties"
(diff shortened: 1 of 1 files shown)


davidhunter08:
I like the second option but maybe move the last part of the line to the 'Using call to action and links' section.

For example:

On to the "do not" list in 'How to use':

> - use breadcrumbs or a standard back link at the top

 In the 'Using call to action and links' section (as a third bullet point in the 'Use a' list):

> - text link to navigate users back to a specific area of the app, for example, "Go to appointments"

Then in GitHub we can log the reasoning for not using the standard back link?

davidhunter08:
I've updated the links to say 'Go to appointments', 'Go to home'.

Graham-Pembrey:
Agree with your suggestions here.

davidhunter08 (approved):

PR #330, opened 2025-06-27
Add patterns section

davidhunter08:
## Todo

- [ ] add patterns link to main nav
- [ ] add patterns section to homepage

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -45,6 +45,11 @@ export default function (eleventyConfig) {
     'node_modules/nhsuk-frontend/packages': 'nhsuk-frontend'
   })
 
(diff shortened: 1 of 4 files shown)


Graham-Pembrey:
Looks good. Wondering whether or not we should include a second paragraph like this, as GOV.UK do. Would it possibly help people understand the relationship between patterns and components?

> Patterns often use one or more [components ](https://design-system.nhsapp.service.nhs.uk/components/)and explain how to adapt them to the context.

davidhunter08:
> Patterns often use one or more [components ](https://design-system.nhsapp.service.nhs.uk/components/)and explain how to adapt them to the context.

Agree. I've updated the content.

PR #399, opened 2025-09-16
Deprecate section heading component

davidhunter08:
## What

Removing the section heading component from the NHS App frontend.

'Archiving' the section heading guidance, with added information on what to use instead.

Screenshot 2025-09-17 at 10 55 07

## Why

The latest nav team design changes include removing the "view all" link from the section heading.

| Before | After |
| -------|-------|
| Image | Image |

This means we can remove the custom mark-up / styles needed for the "view all" link and use standard NHS design system [typography](https://service-manual.nhs.uk/design-system/styles/typography) and [spacing](https://service-manual.nhs.uk/design-system/styles/spacing) instead.

Also, [updates to the card link component](https://github.com/nhsuk/nhsapp-frontend/pull/382) include adding an option for a heading and description for groups of cards. 

## To do

- [x] Remove section heading code
- [x] Check and update all guidance referencing the section heading

### DIFF
--- a/docs/patterns/homepage.md
+++ b/docs/patterns/homepage.md
@@ -16,10 +16,6 @@ This main heading includes the name of the user and a greeting that changes depe
 
 Below the name, we display the NHS number of the user. In research, participants regularly tell us they find having their NHS number here to be useful.
 
-### Section heading
-
-We use [section headings](/components/section-heading/) to separate links by which hub page of the app they come from. Alongside these section headings, ‘View all’ links lead to the relevant hub page.
-
 ### Card links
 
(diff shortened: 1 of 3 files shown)


davidhunter08:
If we update the NHS App with these changes before implementing the new IA, then the 'view all' links will be removed from the homepage, is that ok? Considering the bottom nav can be used to navigate to those sections.

| Before | After |
| -------|-------|
| Screenshot 2025-09-16 at 11 13 10 | Screenshot 2025-09-16 at 11 15 21 |

Tosin-Balogun (approved):
Looks good to me 👍🏿

PR #405, opened 2025-09-26
Update card link guidance

Graham-Pembrey:
- Add guidance for new variants (for campaign and account card)
- Remove card link with an icon (as not used on the app anymore)
- Add card link accessibility guidance
- Add new 'Help users to learn about health campaigns' (related to the campaign card)


Graham-Pembrey:
I'm also sharing this with Maria and Dan Collins, for their view on the new campaign pattern guidance.

davidhunter08 (approved):
LGTM

We can do the final content checks on the v5.0.0 branch.

PR #438, opened 2026-03-19
Feature/ia guidance

zachrigby-nhs:
Hey @michaelgallagher @davidhunter08 

This is the first draft of the IA guidance on the design system site. Still some tweaks to be made such as imagery updates, so its not quite ready to merge yet but raising this PR for you to have a look if you fancy

A couple things to note:
- I added a template property (?) which allows the pages to be ordered in the sidebar based on a numerical value, rather than being arranged alphabetically. I thought this made sense but let me know what you think
- I worked with Ed to add an extra lint fix command after some of the lint tests failed


Graham-Pembrey:
These are looking great! I've started making some light edits to the structure, hub and sub-hub pages. See what you think. A few other thoughts for discussion:

**Overlap between the structure and hub pages**
There is some overlap between the content in the 'Structure' page and the 'Hub' page, where we describe what the three main hub pages do. Are we ok with this? Or should we briefly name the hub pages on the 'Structure' page before signposting to the 'Hub' one for more detail?

**Pages vs screens**
We've used the term 'page' quite a lot, e.g. hub page, sub-hub page. This absolutely fits with the rest of the design system (e.g. error pages) but in the Native Transformation team there has been quite a lot of chat about moving towards using the word 'screen' which has closer connotations with apps, whereas 'page' is more of a website term. We already refer to the home screen on NHS App help pages and in the latest onboarding screens. As we're moving more towards a native app I'm wondering whether to work this change in. Any thoughts? The design system section is called 'Pages' though and it's baked into quite a lot of places on the design system, so maybe this is something for us to consider separately... 

**Referring to our internal workings**
The guidance mentions the Navigation and Onboarding Team in a few places and our internal processes for making changes to the structure. How do we feel about discussing internal details like this on design system pages? Gov.uk for example don't seem to do this, presumably because they have external organisations in mind as a key audience. Should we do the same and leave out internal organisational references? Again this is maybe a bigger question because there are other design system pages where we do this.

Graham-Pembrey (approved):
Looks good to me! I've made a few further small edits. The notes I left above in the PR might be worth considering for future updates.

michaelgallagher:
@zachrigby-nhs one thought: 

in the section listing "when proposing change...", would it make sense to explicitly call out something along the lines of "whether your service could be integrated into an existing service"

thinking about nudging people to go beyond just thinking about where you place a link, but also whether there thing really needs to be a discrete thing at all.

michaelgallagher (approved):
yep, looks good

Issue #450, opened 2026-05-06
App integration pattern for 3rd party sites and services


markasrobinson:
## The need

We believe there may be a need for guidance in the NHS App design system that outlines how 3rd party sites and services that integrate with the app should look and work – an “app integration pattern”.

## What's the problem?

During some work in late 2025/early 2026, we discovered that NHS website content in the app was integrated in inconsistent ways. 

Some pages opened the full website in an overlay, some pages opened the full website in a web view, and some pages opened a slightly customised (with a suppressed header) version of the site in a web view.

This causes a range of issues for users, such as:

- integration patterns can change mid-way through a journey, leading to a jarring and confusing experience
- standard website navigation elements can lead to circular journeys (for example, if users click on “My account” in the website header when they’re in the app, they’ll be prompted to go through the NHS login journey just to end up back on the app)
- misleading or unhelpful terminology is shown to users (for example, links saying “NHS App” take users to pages about the NHS App rather than to the app itself and links to “profile editor” can be interpreted as being related to the user’s app profile rather than profile pages for services on the NHS website
- the search bar searches the NHS website, not the app, and returns results that can be misleading in an app context
- breadcrumbs take users up a level in the website hierarchy, rather than back to the previous page, causing confusing journeys and ultimately trapping users within NHS website pages when they access them from the app 
- lots of distracting links in the website header and footer can take users away from the task they came to the app to complete

The inconsistency also causes some issues for the NHS website, such as:

- analytics data is easier to gather when content is consistently integrated in a web view
- journeys that involve passing across patient data work appear to more smoothy in a web view
- the general direction of travel for app integrations in the future is a “custom overlay”, which is a variation on the current web view integration (not the standard overlay pattern currently used for a lot of NHS website pages)

## Our recommendations for the NHS website

Our team has recommended that NHS website pages in the app should always be displayed using a web view rather than an overlay, and that NHS website teams should implement a consistent “app integration pattern” that customises global navigation elements when we detect a user is accessing the content via the NHS App.

The integration pattern consists of 3 main adaptations for app users:

1. Suppressed site header – total suppression of the NHS website header, including NHS logo, account button, search bar and site navigation links
2. Simplified site footer – a reduction in the number of footer links displayed so that only essential links (terms and conditions, cookies policy and accessibility statement) are shown
3. Backwards navigation – a mechanism for allowing users to return to the previous page in their browser history (generally this means replacing the standard site breadcrumb with a “Back” link)

[...]

PR #457, opened 2026-05-22
Add 'Access web journeys in the native NHS App' pattern guidance

davidhunter08:
## Description

Add 'Access web journeys in the native NHS App' pattern guidance.

## To do

- [x] Add guidance to 'Get started' section  - Dave
- [x] Draft guidance - Graham
- [x] Add images/gifs - Dave
- [x] Review guidance and share with Native Transformation team - Dave/Graham
- [x] Review guidance with NHS App teams - Dave/Graham
- [x] Revise guidance
- [ ] Publish guidance

## Related

Task:
- [x] [Prototype nhs.uk journeys using custom web view](https://github.com/NHSDigital/native-nhsapp-ios-prototype/issues/9) - Dave

Future plan: 
- [x] Merge guidance into a more rounded "Access website journeys through the NHS App" design pattern for all teams internal/external - Dave/Graham


davidhunter08:
On journeys like Patchs and 111 Online, we use the custom web view, but we can't add the close with an alert or the done button because we don't have control  of those URLs so we can't be certain of which URLs to place them on. I'm wondering if we need to explicitly say something about this in the guidance? 

For example:
> On authenticated NHS website journeys (such as 111 online) and third-party journeys (such as Patchs), use the X icon on every page. Do not use the exit alert or a Done button on these journeys. Because you do not control these pages, you cannot reliably tell which page has data input or which is the final page, so neither control can be placed accurately.

What do you reckon? @Graham-Pembrey @michaelgallagher

Graham-Pembrey:
That makes sense and reads well to me 👍  Small possible tweak: should we use "X button" instead of "X icon"?

Graham-Pembrey:
Notes from a call with the cross-domain experience (XDX) team:

Everyone agreed that we should aim to use the custom web view for NHS.UK journeys, whether or not they are authenticated. This fits with the [NHS.UK team recommendations](https://github.com/nhsuk/nhsapp-frontend/issues/450) and evidence they have gathered.  

Some questions we need to work through:

- How should the custom web view back button work on NHS.UK pages? There are various potential entry and exit points. For example, the Health A to Z home page might be the start of some journeys, but in the middle of others. 

- How much are able to hide the footer on NHS.UK pages? There is an NHS.UK requirement to give access to the privacy policy, for example.

- If a user is in a custom web view and opens a third-party link (there are lots on the NHS website) would that open in a custom web view? Or can we set some rules to make sure third-party links jump into a standard web browser, e.g. by doing this when a non-NHS.UK URL is detected? 

Another general point from the call: we might need to reframe the guidance to be more about 'here's what to do if you need to modify the content being shown in an overlay' and 'here's what to do if you don't'.

docs/patterns/access-web-journeys.md
```diff
@@ -0,0 +1,166 @@
...
+  Work in progress
+</strong>
+
+How to adapt your web journey so users can access it in the native NHS App, using a web overlay.
+
+## Overview
+
+For the first native NHS App release, users will start their journeys on native menu screens. If they then select a website-hosted NHS App service from one of those menu screens, it will open in a web overlay.
+
+<figure class="nhsuk-image">
```
Graham-Pembrey, line 21: @davidhunter08 when you have time, would you be able to update this video so the app help journey (around 7 seconds) opens in the latest way?

davidhunter08: done 👍

Graham-Pembrey: Thanks Dave!

Graham-Pembrey:
@michaelgallagher @Tosin-Balogun @frankieroberto I've added you as reviewers, as we think this is at a stage where we can open it up for any feedback or refinements before publishing. No rush though!

michaelgallagher:
three notes: 

1. i think we could clarify that the sheet used to access nhs pages is also a custom webview
2. are we definitely putting the help component at the bottom of custom web views? 
3. we should add some android examples (but this can happen later)

davidhunter08:
> 2\. are we definitely putting the help component at the bottom of custom web views?

I don't _think_ we have settled on this yet, could do with chatting about it.

davidhunter08:
> 3\. we should add some android examples (but this can happen later)

💯

Graham-Pembrey:
> 2. are we definitely putting the help component at the bottom of custom web views?

> I don't think we have settled on this yet, could do with chatting about it.

Agreed, I think this is still very up for discussion!

> 1. i think we could clarify that the sheet used to access nhs pages is also a custom webview

Is that covered by the last sentence below? Maybe we need to move that point or make it stand out more?

image

Graham-Pembrey:
Some quick thoughts on options for the app help footer:

1. Keep it there at all times during NHS App journeys (repeat prescriptions, appointments, test results etc). If users need help part way through a journey but don't want to abandon what they're doing, this gives them a way to do that. 

2. Only use the footer for transactional journeys (so not static ones like test results). Users can easily return to the menu screens from static journeys, where they can access help.

3. Leave it off entirely. Users can close the overlay and access help from any of the home or hub screens. This might reduce the cognitive load as they complete journeys, as it means one less thing on the screen to think about. Also, our help content won't necessarily cover very specific questions users may have about a form they're completing; it's more of a general resource, so maybe this option reflects that better.

4. Base it on the particular journey, and leave it to the discretion of teams. So making it an optional part of the full-screen custom web view. (Not sure this is the right approach as it could lead to lots of inconsistency, but throwing it out there). 

Thoughts welcome! I'll also check with Daniel and Liza whether the research findings can help us decide....

Graham-Pembrey:
From Daniel on the above:

> From the research that we conducted, we didn't observe any users clicking or tapping into the app held within journeys. When we asked where they expected to find the section, everyone went back to the home page, the bottom of the home page, or profile. We didn't test the journeys with real data or in real scenarios, in real life people might try to look for help within the journeys, but we don't know it for sure.
> 
> It may be worth checking with the Analytics team whether people are using the app help in the current app within the journeys though not sure we can get data on that?
> 
> We'd probably lean towards keeping it in the same locations as where it is in the current app and maybe see if we can observe behavior when the native app is live or through moderated live observation.

michaelgallagher:
re: sheet + custom webview 

> Is that covered by the last sentence below? Maybe we need to move that point or make it stand out more?

i don't think it quite does. there is a distinction between nhs.uk journeys and the technology. we might also open bits of the app itself this way, or 111OL, etc.

Graham-Pembrey:
Thanks @michaelgallagher. I've updated that, but it might need further tweaks. See what you think.

michaelgallagher:
@Graham-Pembrey that reads well to me. 

one other thing: 

hiding the page header (i.e. the blue NHS nav element) is generally a good thing for reducing confusion about where the person is, etc. 

i wonder if we want to promote this: 

> Here, the NHS website header (including the search bar) and the breadcrumb links at the top of the page have been hidden. This prevents users from navigating to other parts of the NHS website and getting lost, which we've seen happen often in past user research.

into the main content body, to make it more obvious? 

thinking specifically about 3rd-party suppliers, and how this can be a point of contention, but we **know** having an app/site header in the overlay is confusing. 

so, perhaps putting a finer point on it.

frankieroberto (approved):
Looks good to me.

Would be nice to add some code examples of how to implement the patten (in iOS and Android) - but that can come later.

PR #465, opened 2026-06-09
Add initial page for iOS

frankieroberto:
This is intended as a place where we can start to document design related things for the iOS native app.

It’s an orphaned page for now, as we build up the content.

It can also later be split into sub-pages for components and colours and other things.


michaelgallagher (commented):
looks good to me. @davidhunter08 happy with this? 
if so, i can make a page for android.

davidhunter08 (approved):
🚀

PR #479, opened 2026-06-10
Update links to NHS prototype kit site

frankieroberto:
Some of these URLs have been renamed (we changed `/how-tos/` to `/guides`) so this updates them.

One of the links - to `/install/simple` - was also broken so this fixes that.


davidhunter08 (approved):
Thanks @frankieroberto

PR #480, opened 2026-06-10
Add the initial Android page

michaelgallagher:
An unlinked stub for documenting Android things in the native app. We can use it to describe the approach + scope. 

It can also later be split into sub-pages for components and colours and other things.

### DIFF
--- a/docs/android/index.md
+++ b/docs/android/index.md
@@ -1,0 +1,40 @@
+---
+layout: layouts/base.njk
+title: Android
+description: TODO
+---
+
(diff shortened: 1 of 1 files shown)


frankieroberto:
See #465 for the iOS equivalent.

docs/android/index.md
```diff
@@ -0,0 +1,41 @@
+---
+layout: layouts/base.njk
+title: Android
+description: TODO
```
frankieroberto: Drop description for now?

```suggestion
```

frankieroberto (approved):

PR #486, opened 2026-06-11
Add iOS colour reference

frankieroberto:
This adds an initial reference page for the iOS colours.

I used copilot to read the colours from https://github.com/NHSDigital/nhsapp-design-system-ios and copy the names and hexcodes into this page, which worked ok as a one-off but might need a way of keeping it up-to-date somehow, as these colours are bound to change.

This also reveals a couple of bugs in the the colours list itself - `nhsWarmYellow` and `prescriptionGreen` are missing the dark mode variants somehow.

### DIFF
--- a/docs/_includes/layouts/ios.njk
+++ b/docs/_includes/layouts/ios.njk
@@ -1,0 +1,54 @@
+{% extends "./sidebar.njk" %}
+{% from 'nhsapp/components/tag/macro.njk' import nhsappTag %}
+
+{%- from "./partials/side-navigation.njk" import appSideNavigation %}
+
+{% block beforeContent %}
+  {% if title === "iOS" %}
+    {{ breadcrumb({
+      href: "/",
+      text: "Home"
+    }) }}
+  {% else %}
+    {{ breadcrumb({
+      items: [
+        {
+          href: "/",
+          text: "Home"
+        }
+      ],
+      href: "/ios",
+      text: "iOS"
+    }) }}
+  {% endif %}
+{% endblock %}
+
+{% set sidebar %}
+  {{ appSideNavigation({
+    currentPath: page.url,
+    sections: [
+      {
+        items: collections.ios | sort(attribute="data.title")
+      }
+    ]
+  }) }}
+{% endset %}
+
+{% block content %}
+    {{ nhsappTag({
+    text: 'Work in progress',
+    classes: 'nhsapp-tag--blue nhsuk-u-margin-bottom-4'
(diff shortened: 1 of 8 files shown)


davidhunter08:
`prescriptionGreen` and `prescriptionWhite` can be removed, both are legacy colours from early native prototypes 👍

davidhunter08:
A dark mode variant for `nhsWarmYellow` hasn't been defined yet. It's only currently used for the profile banner for switched profiles and stays the same colour in light and dark mode.

davidhunter08:
Would a width % on the table `<th>` and `<td>`'s fix the alignment?

Screenshot 2026-06-12 at 09 18 47

davidhunter08:
I know we agreed not to have single source of truth for design tokens across the two platforms (iOS and Android) but would it also be useful for the design system docs (and also Figma 🙊)?

Going one step further, could there be a single source of truth for all colour tokens (web and native)? Considering we want to add dark mode colours (and the NHS App pale and dark colour variants) to the `nhsuk-frontend` at some point.

Something to think about?

davidhunter08:
This is something I played about with 8 months ago but didn't get very far https://github.com/davidhunter08/nhsapp-design-tokens - using https://styledictionary.com/

frankieroberto:
Personally I think it’s probably too early to worry about defining colours in a cross platform way at the mo. They might end up diverging in platform-specific ways, particularly if we end up adding more semantic colours. (Or they might not).

For web we might want to upstream the dark mode colours to NHS frontend rather than NHS app frontend?

Worth keeping an eye on though - if it becomes a pain to manage we can add some cross-platform tooling?

davidhunter08:
Fair point, it is too early 👍

frankieroberto:
@davidhunter08 I think I've fixed the table layout, dropped the 2 prescription colours, and tweaked the content a bit. Mind re-revieiwng when you have a mo?

davidhunter08:
Nice! Looking good.

Only minor quibble is the `.nhsGrey5` colour dot blending into the page background:

Screenshot 2026-06-15 at 14 32 37
 
 I don't think it can be targeted and a border applied? It's not a showstopper!

davidhunter08 (approved):
🎨 🚀

frankieroberto:
@davidhunter08 good spot, fixed:

Screenshot 2026-06-15 at 14 46 55

(Using the same pattern as on https://service-manual.nhs.uk/design-system/styles/colour )

PR #489, opened 2026-06-16
Add details on how to set up global tint colour

frankieroberto:
This adds some instructions to the `/ios/colour` page about how to set up the NHS brand colour as the global tint colour.

This then gets inherited by all views, and is used by some native SwiftUI components.

It can't be done automatically by the Swift package so needs a 1-line bit of config, it seems.

Screenshot 2026-06-16 at 15 45 08


davidhunter08:
Thanks for adding this! I've been implementing this in the prototype and have some feedback:

**1. Naming mismatch**

The guidance shows `.tint(.nhsAccent)` but the iOS package currently exposes `.nhsAccentColor`. Either the package needs an `nhsAccent` alias or the docs need updating to match the actual property name.

**2. Missing step — Xcode build warning**

Adding `.tint(.nhsAccentColor)` alone doesn't fully remove the Xcode build warning. You also need to remove the `ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor` build setting from the project (Xcode → Build Settings → search "accent") and delete any `AccentColor.colorset` from your asset catalog. The guidance should mention this step — we verified it clears the warning when done.

**3. Unintended impact on existing icon colours**

Applying `.tint()` at the root `WindowGroup` propagates to every tintable SwiftUI element in the view hierarchy — not just the components you'd expect like toggles. In the prototype, this changed several toolbar icons from black to blue, including the messages icon, flag icon, and trash icon, where no explicit colour had previously been set.

**Design question:** Should toolbar action icons (flag, remove, navigation) be blue by design?

| Before | After |
| -------|-------|
| Simulator Screenshot - iPhone 17 Pro Max - 2026-06-17 at 10 07 35 | Simulator Screenshot - iPhone 17 Pro Max - 2026-06-17 at 10 06 50 |
|  Simulator Screenshot - iPhone 17 Pro Max - 2026-06-17 at 10 07 30 | Simulator Screenshot - iPhone 17 Pro Max - 2026-06-17 at 10 06 56 |

Thoughts @michaelgallagher ?

frankieroberto:
@davidhunter08 good feedback! 

Colour name fixed in https://github.com/NHSDigital/nhsapp-design-system-ios/pull/33 🤦 

Have updated this PR to try and explain that teams will remove the `AccentColor` from their assets catalog to avoid the conflict. Apple seems to add a placeholder colour by default.

Not sure if it needs more detailed guidance on _how_ to do this or not? There are 2 ways I think, via the assets calatalog interface, or by editing the code behind that.

Screenshot 2026-06-17 at 10 59 00

frankieroberto:
I did a quick exploration of which SwiftUI native views use the `tint()` color by default:

Screenshot 2026-06-17 at 11 01 16

I'm also undecided on whether we'd want all of these to have the tint colour, or only some of them. 

For any of our custom components we could decide on a case by case basis - but there might be times when the raw native components are still used?

davidhunter08:
The only other components I've seen it change the colour in the prototype for are `.alert` (example below) and `.toolbar` (example above comment) buttons.

| Before | After |
| -------|-------|
| Simulator Screenshot - iPhone 17 Pro Max - 2026-06-17 at 11 36 16 | Simulator Screenshot - iPhone 17 Pro Max - 2026-06-17 at 10 00 44 |

It feels like setting the `.tint` colour is the right thing to do, so all default iOS use it, and then like you said, make overrides where needed?

frankieroberto:
Just discovered that toolbar buttons use the tint by default too.

Apple's own apps don't seem to use a tint for these though (despite there being a default tint colour) 🤔 

Makes me think that setting up a global tint in your your entry point file is the wrong thing to do...

Screenshot 2026-06-17 at 11 54 01

michaelgallagher:
> **Design question:** Should toolbar action icons (flag, remove, navigation) be blue by design?

@davidhunter08 i'm ... not sure. i think it would be fine if they were blue if the back button was also blue, but my initial reaction is that they look weird when the top toolbar/nav also has a black back icon.

davidhunter08:
> > **Design question:** Should toolbar action icons (flag, remove, navigation) be blue by design?
> 
> @davidhunter08 i'm ... not sure. i think it would be fine if they were blue if the back button was also blue, but my initial reaction is that they look weird when the top toolbar/nav also has a black back icon.

FWIW I don't like the blue icons, that would have been one of the components I'd suggest we customise if we were to implement a blue tint globally.

frankieroberto:
Agree - the toolbar icons all being tinted looks weird.

Hmm, so looking into this a bit further, I think there’s a difference between having a color named `AccentColor` in your app's asset catalog, vs adding `.tint()` to your root `ContentView`. (I had initially assumed that the outcome was the same)

Having an `AccentColor` defined affects regular buttons and links, and some other components like Slider, but doesn’t seem to affect other components like Toogles or toolbar buttons. Setting `.tint()` on your root content view on the other hand affects a whole bunch more (basically anything which has a `tint()` option on it?).

In summary, I think our options are:

### 1. don't recommend setting a global `.tint()` or `AccentColor`

This means that toolbar buttons, toggles, and other such native components will remain grey by default.

The only downside is that regular links and buttons will flip back to the default tint colour (which is also blue, but a slightly lighter shade)

### 2. recommend setting a global .`tint()` but override it

We could keep the recommendation to set a global `tint()` but then override it as needed in any custom components, and recommend that users override it themselves if using the plain native components.

### 3. don't recommend setting a global `.tint()` but instead recommend adding an `AccentColor`

We could remove the recommendation to set up a global tint, but instead recommend adding an `AccentColor` asset using the Asset catalog - this would then only affect buttons, links, sliders and possibly a few other components.

The only downside is that this requires manually adding the colour into your asset catalog, as I don't think the `AccentColor` can be defined by a swift package dependency.


---

In short there’s 2 ways to set an accent/tint color, and they do different things. Apple’s [accentColor](https://developer.apple.com/documentation/swiftui/color/accentcolor) and [tint](https://developer.apple.com/documentation/swiftui/view/tint(_:)-23xyq) docs don't really make this clear (or at least not to me).

I think on balance it’s probably best to go with option 1 for now?

frankieroberto:
@davidhunter08 @michaelgallagher any thoughts on this ☝️? https://github.com/nhsuk/nhsapp-frontend/pull/489#issuecomment-4730593629

I think probably 1) is simplest for now, but we could do 3) if we think it’s good to still recommend setting an AccentColor (for links, sliders, and a small number of other default components) we could do that – it just can’t be done from within the design system package.

davidhunter08:
I think we will set the AccentColor in the iOS prototype so that the links, sliders etc use the nhsBlue colour? So I'd recommend option 3 based on that.

PR #490, opened 2026-06-16
Update eleventy

frankieroberto:
Version 3 is well out of alpha now so we can switch to the latest 3.1.6 release.


davidhunter08 (approved):
Nice one!

PR #496, opened 2026-06-25
Add page about Home menu

frankieroberto:
Add an initial page about the first component added.

Wasn't sure how much detail to go in about how to actually use this in code. Probably it's a bit unnecessary for developers, but might be more useful for designers prototyping in code?

I looked at SwiftUI documentation pages like [List](https://developer.apple.com/documentation/SwiftUI/List) for example (which has lots of example code), but also at the existing [app web components](https://design-system.nhsapp.service.nhs.uk/components/) pages, which are a bit more design-focused.

Feedback welcome!


docs/ios/home-menu.md
```diff
@@ -0,0 +1,147 @@
+---
+layout: layouts/ios.njk
+title: Home menu
+tags:
+  - iosComponents
+---
+
+The home menu lists the main sections of the app.
+
+<img src="/assets/images/ios/HomeMenu.png" width="320">
```
frankieroberto, line 10: Probably ideally we have some kind of custom screenshot viewer widget here, which lets you toggle between different variants like light mode/dark mode and text sizes?

docs/ios/home-menu.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Accessibility
+
+This component supports Dynamic Type, Dark Mode and VoiceOver.
+
+The `title` for each menu item is read out by VoiceOver. The icon images are not read out.
+
+## Research
+
+This component is not yet being used by the live NHS App, but several rounds of research have been done on it.
```
frankieroberto, line 59: This probably needs a bit more detail?

docs/ios/home-menu.md
frankieroberto: @michaelgallagher hopefully your tool, or something similar, could auto-generate these from the preview canvases at some point? 

Or if it needs to be run in the simulator we can set up example views within a test app in the repo like in https://github.com/NHSDigital/nhsapp-design-system-ios/pull/55

davidhunter08:
An idea to extend the code snippet example https://github.com/nhsuk/nhsapp-frontend/pull/497 - laying the foundations for adding Android (Kotlin) examples in the future 👀🤖

docs/ios/home-menu.md
michaelgallagher: ideally, yes. might make sense to reappraise the tool a little after the DS is settled, as it probably needs some changes to work correctly

michaelgallagher (approved):
this looks good. happy to merge this. 

1 thing for the future: reading this, i think it might be better to cover both iOS and android in one page, showing examples for each. the component will be largely the same, after all.

frankieroberto:
@michaelgallagher let's come back to that tricky question. I suspect we'll end up with some components that are very similar across platforms (except for code implementation) and others that are quite different (different names, behaviour, etc). Not sure how best to handle this from an IA perspective yet, as there's probably needs to navigate by _both_ platform and by component/pattern...

docs/ios/home-menu.md
frankieroberto: We can come back to this later.

PR #497, opened 2026-06-26
Extend example shortcode with Swift code tabs

davidhunter08:
## Summary

- Extends the `{% example %}` shortcode with new frontmatter options: `swiftCode`, `swiftArguments`, `showHtml`, `openFirst`, and `androidCode` (for future use)
- Adds a Swift tab (and Android tab skeleton) to the example tab UI, conditionally shown when the relevant frontmatter is set
- Adds `showHtml: false` to hide the iframe preview and HTML/Nunjucks tabs for native-only components, with a CSS fix to restore the missing top border
- Adds `openFirst: true` to auto-open the first tab on load
- Adds `swiftArguments` to show a collapsible Swift options table, mirroring the existing Nunjucks macro options pattern
- Applies all of the above to the iOS home menu page, replacing raw Swift code blocks with `{% example %}` shortcodes for all three examples (basic usage, navigation, modal sheet)

### Before
Screenshot 2026-06-26 at 16 12 04

### After

Screenshot 2026-06-26 at 16 11 28 
Screenshot 2026-06-26 at 16 11 37

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -123,10 +123,16 @@ export default function (eleventyConfig) {
       showNunjucksAuto = data.showNunjucks
     }
 
+    // Always show HTML preview/tab unless explicitly disabled
+    let showHtmlAuto = true
+    if (typeof data.showHtml === 'boolean') {
+      showHtmlAuto = data.showHtml
+    }
+
     const rawHtmlCode = nunjucksEnv.renderString(nunjucksCode)
-    const prettyHtmlCode = await prettier.format(rawHtmlCode, {
-      parser: 'html'
-    })
+    const prettyHtmlCode = rawHtmlCode.trim()
+      ? await prettier.format(rawHtmlCode, { parser: 'html' })
+      : ''
 
     const href = `/examples/${examplePath.replace('.njk', '')}`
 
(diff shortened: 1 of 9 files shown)


frankieroberto:
Ooh, love this! The code seems to have lost its syntax highlighting though?

The code block with the “copy code” button is now included within NHS frontend so we can also switch to that at some point too, to save maintaining it here. See https://github.com/nhsuk/nhsuk-frontend/pull/1841

davidhunter08:
> The code seems to have lost its syntax highlighting though?

The syntax highlighting is working correctly. The issue I'm seeing locally is caused by a Chrome extension conflicting with the Prism token spans. Confirmed working correctly in incognito mode (and with Safari and Firefox).

Screenshot 2026-06-26 at 20 46 30

davidhunter08:
> The code block with the “copy code” button is now included within NHS frontend so we can also switch to that at some point too, to save maintaining it here. See https://github.com/nhsuk/nhsuk-frontend/pull/1841

Ace! I'm probably best making a separate issue for this?

frankieroberto:
> > The code seems to have lost its syntax highlighting though?
> 
> The syntax highlighting is working correctly. The issue I'm seeing locally is caused by a Chrome extension conflicting with the Prism token spans. Confirmed working correctly in incognito mode (and with Safari and Firefox).

Ah, ace! I didn’t check, sorry - just went off the screenshots.

let’s merge this and make any improvements separately.

frankieroberto (approved):

PR #502, opened 2026-06-29
Add some initial guidance on 3 core scrollable view types

frankieroberto:
This tries to take the research and findings from #44 into the differences between `List`, `Form` and `ScrollView` containers and extract it into some initial guidance pages.

Possibly the thing this is missing though is a parent page which contains a bit of the decision tree from https://github.com/NHSDigital/native-nhsapp-ios-prototype/issues/44#issuecomment-4833073561 and then links to 3 pages for more details? Maybe that hints there should be a generic `/ios/patterns` index page?

Or is 'patterns' even the right term here?

### DIFF
--- a/docs/examples/forms/example.njk
+++ b/docs/examples/forms/example.njk
@@ -1,0 +1,32 @@
+---
+title: Forms
+showHtml: false
+openFirst: true
+swiftCode: |
+  let ratings = [
+    "Very good",
+    "Good",
+    "Neither good nor poor",
+    "Poor",
+    "Very poor"
+  ]
+
+  var body: some View {
+      VStack {
+          Form {
+              Section {
+                  Picker("Rating", selection: $rating) {
+                      ForEach(ratings, id: \.self) { option in
+                          Text(option)
+                      }
+                  }
+                  .pickerStyle(.inline)
+              } header: {
+                  Text("Overall, how would you rate your experience of the service?")
+              }
+          }
+
+          Button("Continue")
+      }
+  }
+---
--- a/docs/_includes/layouts/ios.njk
+++ b/docs/_includes/layouts/ios.njk
@@ -44,6 +44,12 @@
           text: 'Components'
         },
         items: collections.iosComponents | sort(attribute="data.title")
+      },
+      {
+        heading: {
+          text: 'Patterns'
+        },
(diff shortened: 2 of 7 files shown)


docs/examples/forms/example.njk
```diff
@@ -0,0 +1,32 @@
...
+                          Text(option)
+                      }
+                  }
+                  .pickerStyle(.inline)
+              } header: {
+                  Text("Overall, how would you rate your experience of the service?")
+              }
+          }
+
+          Button("Continue")
```
frankieroberto, line 29: These are all unstyled, vanilla SwiftUI components for now - but we can replace them with NHS styled alternatives as they’re added?

davidhunter08:
Love it!

Should we link to the Apple guidance:
- https://developer.apple.com/documentation/swiftui/form
- https://developer.apple.com/documentation/SwiftUI/ScrollView
- https://developer.apple.com/documentation/SwiftUI/List

Or is it too techy?!

davidhunter08:
'Layouts' might be better than 'Patterns'?

frankieroberto:
> Love it!
> 
> Should we link to the Apple guidance:
> 
> * https://developer.apple.com/documentation/swiftui/form
> * https://developer.apple.com/documentation/SwiftUI/ScrollView
> * https://developer.apple.com/documentation/SwiftUI/List
> 
> Or is it too techy?!

There's links already! Maybe they should be more prominent though, if you didn't spot them!

frankieroberto:
> 'Layouts' might be better than 'Patterns'?

Good shout. Done in 8b569b3821626ffdd276e9b1f34561fb0cfcc971

davidhunter08:
> There's links already! Maybe they should be more prominent though, if you didn't spot them!

aha - I think my expectation was the link would be at the top in the description BUT I think it makes more sense for it to be at the bottom 👍

davidhunter08 (approved):
I think it's good for now. I think some screenshots will enhance the guidance but we can add them later, maybe at the same time as updating the code examples to use NHS styled components?

frankieroberto:
@davidhunter08 ok cool, will merge. Feel free to update it as your investigations into List vs Form vs ScrollView continue! 😄

PR #504, opened 2026-07-01
Add guidance on embedding design system package

davidhunter08:
Adds an extra iOS setup check to prevent runtime crashes when consuming the design system Swift package in Xcode, aligning the docs with the embedding requirement raised in https://github.com/NHSDigital/native-nhsapp-ios-prototype/issues/52.


frankieroberto (approved):
Ah nice! I hadn't encountered this as I’ve not tried putting it on a real phone yet (as I don't have developer account setup yet)

PR #505, opened 2026-07-01
Add guidance for spotting when a screen belongs in a scroll view

davidhunter08:
The current guidance says "screens mixing different item types including headings, links, and buttons" but doesn't help someone recognise when they've drifted into the wrong template. 

Line added:

> If you find yourself removing list styling — for example, using clear backgrounds or hiding row separators — that is a sign the screen belongs in a scroll view.


frankieroberto (approved):

frankieroberto:
❤️ it

PR #506, opened 2026-07-01
Add guidance for choosing between form and list views

davidhunter08:
Adds a sentence to the "When to use" section clarifying that screens with a fixed set of input controls — such as toggle rows — belong in a form view, not a list view. 

Also fixes a broken link in "When not to use" that pointed to `/ios/forms` instead of `/ios/scroll-views.`

### DIFF
--- a/docs/ios/forms.md
+++ b/docs/ios/forms.md
@@ -13,11 +13,13 @@ Use a form view when the screen collects user input or configuration, for exampl
 
 A form may contain several items, but the items do not change or get reordered, instead the user changes the values within them.
 
+This includes screens that look like a list, for example a set of toggle rows for a multiple-choice question. If the rows are fixed and not backed by data, use a form view.
+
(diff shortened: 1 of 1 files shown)


docs/ios/forms.md
```diff
@@ -13,11 +13,13 @@ Use a form view when the screen collects user input or configuration, for exampl
 
 A form may contain several items, but the items do not change or get reordered, instead the user changes the values within them.
 
+This includes screens that look like a list, for example a set of toggle rows for a multiple-choice question. If the rows are fixed and not backed by data, use a form view.
```
frankieroberto: Avoid starting a para with "This"?

```suggestion
Forms can include screens that look like a list, for example a set of toggle rows for a multiple-choice question. If the rows are fixed and not backed by data, use a form view.
```

frankieroberto (approved):

PR #515, opened 2026-07-01
Upgrade Node from v20 to v24

frankieroberto:
This matches NHS frontend. See https://github.com/nhsuk/nhsuk-frontend/pull/1673


colinrotherham (approved):
Lots of packages need updating, do they work fully with Node.js 24?

frankieroberto:
@colinrotherham seem to!

PR #516, opened 2026-07-01
Switch to Sass modules for docs site

frankieroberto:
Switch the docs site's Sass compilation from the legacy @import API to the Sass module system.


.eleventy.js
```diff
@@ -83,13 +83,15 @@ export default function (eleventyConfig) {
         return
       }
       let result = sass.compileString(inputContent, {
-        // Expanded load paths so @import "nhsuk/index" and other bare imports resolve
+        // Expanded load paths so bare imports resolve
         loadPaths: [
           '.',
           'node_modules',
           'node_modules/nhsuk-frontend/dist',
           'node_modules/nhsuk-frontend/src'
```
colinrotherham: We should remove src

.eleventy.js
```diff
@@ -83,13 +83,15 @@ export default function (eleventyConfig) {
         return
       }
       let result = sass.compileString(inputContent, {
-        // Expanded load paths so @import "nhsuk/index" and other bare imports resolve
+        // Expanded load paths so bare imports resolve
         loadPaths: [
           '.',
```
colinrotherham: We should remove `.` otherwise local files are treated as dependencies

.eleventy.js
```diff
@@ -83,13 +83,15 @@ export default function (eleventyConfig) {
...
           'node_modules',
           'node_modules/nhsuk-frontend/dist',
           'node_modules/nhsuk-frontend/src'
-        ]
+        ],
+        // Required to resolve pkg: URL imports (e.g. @use "pkg:nhsuk-frontend/...")
+        importers: [new sass.NodePackageImporter()]
```
colinrotherham, line 88: Probably want to grab source maps and other bits? https://github.com/nhsuk/nhsuk-frontend/blob/main/shared/tasks/styles.mjs

docs/assets/css/app.scss
```diff
@@ -1,29 +1,29 @@
...
+@use "pkg:nhsuk-frontend/dist/nhsuk" as *;
 
 // Import all styles from the NHS App frontend library
-@import "src/all";
+@forward "src/all";
 
 // Code example syntax highlighting
-@import "node_modules/prismjs/themes/prism";
-@import "docs/assets/css/code-colours";
+@use "node_modules/prismjs/themes/prism.css";
```
colinrotherham: Either use pkg or move the `node_modules` bit to load paths?

colinrotherham (commented):
Can you carefully check the output for uncompiled `nhsuk-*` things? Just in case

docs/assets/css/app.scss
frankieroberto: Done.

The prism styles will be removed in a follow-up PR and replaced with the syntax highlighting styles from NHS frontend.

.eleventy.js
frankieroberto: Done, thanks.

.eleventy.js
frankieroberto: Done, thanks.

.eleventy.js
frankieroberto: I think I’ll leave this for now, if that's ok, and leave this for a follow-up PR?

We might be able to switch this eleventy build to using the NHS eleventy plugin, which will mean that the sass compilation config moves upstream - so sourcemaps could be configured here instead: https://github.com/x-govuk/govuk-eleventy-plugin/blob/main/src/extensions/scss.js

frankieroberto:
> Can you carefully check the output for uncompiled `nhsuk-*` things? Just in case

Just checked (both manually and a few times with AI) and can't find anything.

.eleventy.js
colinrotherham: If you're sure it's not instantly useful as a quick boolean to add?

.eleventy.js
```diff
@@ -82,14 +82,10 @@ export default function (eleventyConfig) {
...
       }
-      let result = sass.compileString(inputContent, {
-        // Expanded load paths so @import "nhsuk/index" and other bare imports resolve
-        loadPaths: [
-          '.',
-          'node_modules',
-          'node_modules/nhsuk-frontend/dist',
-          'node_modules/nhsuk-frontend/src'
-        ]
+      let result = sass.compile(inputPath, {
```
colinrotherham: Should this be ~`let`~ `const`?

We should either remove the `async` further up or make use of the Sass async API here

```suggestion
      const result = sass.compileAsync(inputPath, {
```

colinrotherham (approved):

.eleventy.js
frankieroberto: Will look at this as part of a follow-up

PR #519, opened 2026-07-06
Use code and highlight from NHS frontend

frankieroberto:
A recent version of NHS frontend included styles for syntax highlighting, and a Code component which has an optional 'Copy code' button.

This means we can swap out our custom syntax highlighting and code components, and use the ones upstream. Less for us to maintain, plus we inherit some accessibility improvements and bug fixes too.

I've moved the Swift code examples back into the markdown files that mention them, for now, as they don't really use the tabs interface given there's only a single language and no preview. The "Copy code" buttons still appear though.

There are some visual differences in the syntax highlighting and Copy code buttons - I think @anandamaryon1 worked on making the highlight colours more accessible (higher contrast), and the Copy code buttons uses the secondary button style.

## Screenshots

| Before | After |
|----|-----|
| Screenshot 2026-07-08 at 13 01 57 | Screenshot 2026-07-08 at 13 03 58 |
| Screenshot 2026-07-08 at 13 05 15 | Screenshot 2026-07-08 at 13 04 42 |

The upstream component also introduces a focus style, which is only added if the code doesn't fit in its container and needs to scroll (this is useful for accessibility):

Screenshot 2026-07-08 at 13 21 10


frankieroberto:
@colinrotherham just pointed out that the new code component supports using arrow keys for scrolling too. ⬇️ ➡️  🙌

.eleventy.js
```diff
@@ -158,7 +183,58 @@ export default function (eleventyConfig) {
...
+      let html = `<div class="${codeClasses}" data-module="nhsuk-code">\n`
+
+      // Button is hidden by default and shown by JavaScript when clipboard API is available
+      if (hasCopyButton) {
+        html += `  <button class="${buttonClasses}" data-module="nhsuk-button" type="button" hidden>Copy code</button>\n`
+      }
+
+      html += `  <pre class="nhsuk-code__container"><code class="nhsuk-code__content">${code}</code></pre>\n`
+      html += `</div>\n`
+      return html
```
colinrotherham: Can we render the code component via `nhsuk-frontend/lib`? 🙏

Saves worrying about all these buttons classes and reverse styles

```js
import { components } from 'nhsuk-frontend/lib'

components.render('code', {
  context: {
    html: '<p>This is an HTML code block.</p>',
    background: 'body',
    button: true,
    reverse: true
  }
})
```

frankieroberto: Trying that in this PR: https://github.com/nhsuk/nhsapp-frontend/pull/520

frankieroberto: @colinrotherham helped me fix this, thanks!

frankieroberto:
@davidhunter08 @colinrotherham happy with this being merged?

colinrotherham:
@frankieroberto Not yet, we've pulled out the highlight filter

Might as well use it directly?

```patch
      const language = token.info.trim()
-     const languages = language ? [language] : undefined
-
-     const { value: callBlock } = highlighter.highlightAuto(
-       token.content,
-       languages
-     )
+     const callBlock = highlightFilter(token.content, language)
```

frankieroberto:
> @frankieroberto Not yet, we've pulled out the highlight filter
> 
> Might as well use it directly?

Nice, thanks for pushing this commit! 🙌

davidhunter08:
Nice @frankieroberto and @colinrotherham!

One small thing, the copy button on some components (badge, card links, tag, timeline) for the Nunjucks option doesn't have the correct spacing...

Screenshot 2026-07-13 at 10 31 59

The spacing is ok on other components (buttons and header)...

Screenshot 2026-07-13 at 10 39 26

frankieroberto:
@davidhunter08 good spot! Fixed in 5fb24d546b9db18eef794c24326bbd7d80d6db6a - was caused by Markdown adding a bunch of `<p>` tags.

davidhunter08 (approved):
LGTM!

PR #522, opened 2026-07-15
Add guidance for iOS toolbar components

frankieroberto:
Add guidance for the toolbar items added in https://github.com/NHSDigital/nhsapp-design-system-ios/pull/54

➡️ [Preview](https://deploy-preview-522--nhsapp-frontend.netlify.app/ios/toolbar/)


davidhunter08 (approved):
Great stuff @frankieroberto ❤️

Only thing I'm not sure on is the size of the images, but I think that's something we're going to refactor as a whole later anyway so it's not a blocker.

frankieroberto:
@davidhunter08 yeah, agreed about the images. It's also slightly annoying to have to switch from Markdown to HTML just to set the image width. At full (2/3rds) width they look a bit huge.

frankieroberto:
@Graham-Pembrey I'll merge this now, but if you have any suggestions for improvements we can keep iterating it!

PR #530, opened 2026-07-17
Bugfix: show styles section as active

frankieroberto:
Spotted this minor bug.


davidhunter08 (approved):
Thanks @frankieroberto

PR #532, opened 2026-07-20
Remove unused examples

frankieroberto:
These were referenced from the Homepage and Page not found pages, but they were retired in https://github.com/nhsuk/nhsapp-frontend/pull/531


davidhunter08 (approved):
Thanks @frankieroberto

PR #538, opened 2026-07-24
Integration pattern separated

flora-nhs:
separating the content into 5 distinct pages across Get Started, and Patterns


michaelgallagher:
one quick observation: 

in the "getting started" section, it seems odd to have the section title and the page title be the same

Graham-Pembrey:
Some quick thoughts:

## Getting started section

I agree with Mike about the title of the section and the page being the same. Perhaps the side-panel title could just be 'Integrations'? That matches the one-word titles like 'Prototyping' and 'Production'.

## Patterns section

I think the child pages work here. An alternative might be to have a new side-bar section of 'Integrations' with the new pages under it? Not sure if 'Pages' is the right section title or not?

### Custom web view integration page

- Step 3 in the 'How to adapt your web service' part: we've got this marked as only applying to the current app, not the future one, but I think it would apply to both? We'd still want a minimal footer in the native custom overlay on NHS.UK pages? 
- I'm wondering if we can use the [details component](https://service-manual.nhs.uk/design-system/components/details) to hide all of the 'in the future native NHS App' sections? That could neatly separate the future parts out. At the moment it feels like presenting the current and future together is quite a lot to take in, and it means context switching while reading through. If we made the page about the current app but sectioned out all the future stuff into expandable details bits, people working on the future version could open them as needed? We could also bring the animated gifs from the current future state guidance into these detail components.

## General thought

Do the integrations team (Liz Marler and co) need to review this? I think they own the NHS Digital website guidance we're linking to.

PR #540, opened 2026-07-29
Add default sizes to iOS font table

davidhunter08:
Screenshot 2026-07-29 at 15 53 04

### DIFF
--- a/docs/ios/typography.md
+++ b/docs/ios/typography.md
@@ -29,7 +29,7 @@ The styles are defined relative to the native SwiftUI Font styles - this means t
 | `.nhsCaption`     | `.caption`         | 12pt         |
 | `.nhsCaption2`    | `.caption2`        | 11pt         |
 
-All styles are Regular weight except `.nhsHeadline`, which is Semibold.
+All styles are Regular weight except `.nhsHeadline`, which is Bold.
 
 ## Custom sizes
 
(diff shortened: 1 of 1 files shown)


Tosin-Balogun (approved):
Looks good to me and is quite useful to see the scales. One observation is that `extraLarge` is left out. Is this intentional?

davidhunter08:
> Looks good to me and is quite useful to see the scales. One observation is that `extraLarge` is left out. Is this intentional?

Good spot! They have not been added to the NHS design system, probably because we haven't found a need for them yet but we should probably add them? I'll create an issue to get them added.

davidhunter08:
@Tosin-Balogun I did some digging...

Font sizes for .extraLargeTitle / .extraLargeTitle2

At the default Dynamic Type setting:

- .extraLargeTitle → 36pt, bold
- .extraLargeTitle2 → 28pt, bold

These are semantic text styles, so the values above are defaults — they scale with the user's Dynamic Type / accessibility settings rather than being fixed. Apple's docs don't publish exact point sizes; these are measured/community values.

Note the availability is described inconsistently across sources: some frame these as visionOS-exclusive (introduced for spatial text hierarchy), others list them as available from iOS 17+ as the recommended way to exceed .largeTitle. Worth verifying against your deployment target.

One quirk: .extraLargeTitle (36pt) is only marginally larger than .largeTitle (34pt). If you need something dramatically bigger for a hero heading, .system(size:weight:design:) is usually the better fit — with the tradeoff that it won't scale with Dynamic Type by default.

So I'm not sure if should add them or not?

PR #543, opened 2026-07-31
Add iOS button guidance

davidhunter08:
Related to https://github.com/NHSDigital/nhsapp-design-system-ios/pull/78

[Guidance preview](https://deploy-preview-543--nhsapp-frontend.netlify.app/ios/buttons/)


Tosin-Balogun (commented):
@davidhunter08 Does the catalogue iOS have text button?

davidhunter08:
> @davidhunter08 Does the catalogue iOS have text button?

Do you mean this https://github.com/NHSDigital/nhsapp-design-system-ios/issues/79?

Tosin-Balogun:
> > @davidhunter08 Does the catalogue iOS have text button?
> 
> Do you mean this [NHSDigital/nhsapp-design-system-ios#79]

Yeah, I think it is covered in the example of the 111 online button

`Button("111 online") { }
    .font(.nhsBody)
    .tint(.nhsBlue)`
    
Are you leaving it out of the guidance because it's not ready?

Screenshot 2026-08-05 at 09 55 53

davidhunter08:
> Are you leaving it out of the guidance because it's not ready?

Yeah, until we're clear on how best to implement it I thought it'd best to leave it out for now. What do you reckon?

Tosin-Balogun:
> > Are you leaving it out of the guidance because it's not ready?
> 
> Yeah, until we're clear on how best to implement it I thought it'd best to leave it out for now. What do you reckon?

It is fine if that is the case. Something to consider though is the plain button is actually the first default button type, so component wise, it belongs within this group, perhaps something to update this guidance when its ready?

In Android, currently have these based on Android defaults
Screenshot 2026-08-05 at 10 19 59
Plus the floating action buttons, but going to do an overpass on them once the token work is done

davidhunter08:
> perhaps something to update this guidance when its ready?

I think so 👍

Looking at the Android styles, iOS doesn't have a 'tonal' or 'elevated' version, should they? Are we clear on usage for them? eg: Use a tonal button for X reason.

Tosin-Balogun:
> > perhaps something to update this guidance when its ready?
> 
> I think so 👍
> 
> Looking at the Android styles, iOS doesn't have a 'tonal' or 'elevated' version, should they? Are we clear on usage for them? eg: Use a tonal button for X reason.

Yeah, it is a good point and I do plan to clean it when I do a button pass. I think it is just setting the default button packages first and then extracting the ones being used and not used. Like the family career button used to be in there, but now removed, there's also the toolbar buttons that need factoring in.

I think at the moment, we have not used the tonal or elevated buttons, so will have to mute them when I do a pass

docs/ios/buttons.md
```diff
@@ -0,0 +1,134 @@
...
+    Button("Secondary 2") { }
+}
+.buttonStyle(.nhsSecondary)
+```
+
+## Accessibility
+
+This component supports Dynamic Type, Dark Mode and VoiceOver.
+
+Disabled buttons fade visually but remain in the accessibility tree, so VoiceOver users know the action exists even when it is not available.
```
Graham-Pembrey, line 130: This guidance looks good to me overall! 

Disabled buttons are mentioned here but not elsewhere on the page. They are mentioned in the NHS design system buttons guidance, but as something to generally avoid. Should we give any more context here for when you might include a disabled button or how to implement that on iOS? For example:

> The NHS design system guidance is to only use [disabled buttons](https://service-manual.nhs.uk/design-system/components/buttons#disabled-buttons) when doing so is supported by user research. If you include a disabled button on iOS, it will fade visually but remain in the accessibility tree, so VoiceOver users know the action exists even when it is not available.

frankieroberto: @Graham-Pembrey this might be an area where we deliberately diverge from the NHS design system web guidance - as it’s a much more common pattern on iOS to use disabled buttons that dynamically become enabled when you’re able to press them, rather than having them always enabled but showing an error message if you press them in an invalid state.

I think we’ll probably need a separate 'question pattern' page or similar which gives some guidance on this, and other related issues (like using a modal sheet)?

Shall we create a backlog issue to add this? cc @davidhunter08

frankieroberto (approved):
Nice.

docs/ios/buttons.md
Graham-Pembrey: Sounds good @frankieroberto. I've added a [backlog issue for a question pattern page](https://github.com/orgs/NHSDigital/projects/55/views/1?pane=issue&itemId=233827685&issue=NHSDigital%7Cnative-nhsapp-ucd-team%7C90 ).

I'd be happy to have a look into this. Might put in 15/30 minutes with you and @davidhunter08 this week to chat about what we think it should cover?

PR #551, opened 2026-08-11
NHSO-69998: card link footer redesign

imnatgreen:
- Update the design for stacking card links to support our updated designs for card footers. You can now stack a coloured card link below a complex card to separate related content whilst keeping the overall card easy to scan. This reduces card complexity compared to the existing divider line.

  new
  old

- The clock icon has also been realigned within its bounding box to feel more central when aligned with the small badge when documents and questionnaires are attached:

  

- Added `nhsapp-icon-text` and `nhsapp-badge-small--as-icon` to make the card content layout more reusable.
- Cards now use `first-child` and `last-child` modifiers for borders over `first-of-type` and `last-of-type` to prevent conflicts with unrelated sibling elements.


docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+  classes: '',
+  isListItem: false,
+  cards: [
+    {
+      title: 'Orthopaedic appointment',
+      href: '#',
+      titleClasses: 'nhsuk-u-font-size-22',
+      linkAriaLabel: 'Action needed: Orthopaedic appointment on Monday 3 June 2024 at 9:40am, at The Willows, Croydon University Hospital',
+      descriptionHtml: '
+        <p aria-hidden="true" class="nhsuk-u-font-weight-bold nhsuk-u-margin-top-1 nhsuk-u-margin-bottom-3">Monday 3 June 2024 at 9:40am</p>
```
colinrotherham, line 33: If this is important enough to make visually bold, can we make it an `<h3>`?

Assuming your card heading is an `<h2>`

docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+      '
+    },
+    {
+      classes: 'nhsapp-card--pale-blue',
+      descriptionHtml: '
+        <div class="nhsapp-icon-text">
+          <svg class="nhsapp-icon nhsapp-icon__clock nhsapp-icon--text-colour nhsapp-icon--32" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" height="2rem" width="2rem" viewBox="0 0 24 24">
+            <path d="M23.099 10.701H21.9C21.26 5.796 17.075 2 12 2 6.476 2 2 6.476 2 12c0 5.525 4.476 10 10 10 3.407 0 6.414-1.708 8.222-4.316h-.01c.1-.14.17-.29.17-.47a.77.77 0 0 0-.77-.768c-.25 0-.47.13-.609.32h-.01a8.51 8.51 0 0 1-6.983 3.646 8.5 8.5 0 0 1-8.502-8.502A8.487 8.487 0 0 1 12 3.41c4.286 0 7.812 3.166 8.402 7.292h-1.03c-.21 0-.329.24-.21.41l1.879 2.627c.1.15.32.14.42 0l1.848-2.627a.26.26 0 0 0-.21-.41" />
+            <path d="m14.52 15.251.01.01.04.04c.11.09.25.14.4.14a.644.644 0 0 0 .48-1.069l-.02-.02-.01-.01-3.047-2.887V6.47a.641.641 0 0 0-1.28 0v5.545l3.417 3.246z" />
+          </svg>
```
colinrotherham: I don't suppose you have an app version of the `nhsukIcon()` macro?

```suggestion
          ' ~ nhsukAppIcon("clock") ~ '
```

docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+          <div>
+            <p class="nhsuk-u-font-weight-bold nhsuk-body-m nhsuk-u-margin-bottom-0">Long stay</p>
```
colinrotherham: Should this be a heading too?

docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+    },
+    {
+      classes: 'nhsapp-card--pale-blue',
+      descriptionHtml: '
+        <div class="nhsapp-icon-text">
+          <svg class="nhsapp-icon nhsapp-icon__clock nhsapp-icon--text-colour nhsapp-icon--32" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" height="2rem" width="2rem" viewBox="0 0 24 24">
+            <path d="M23.099 10.701H21.9C21.26 5.796 17.075 2 12 2 6.476 2 2 6.476 2 12c0 5.525 4.476 10 10 10 3.407 0 6.414-1.708 8.222-4.316h-.01c.1-.14.17-.29.17-.47a.77.77 0 0 0-.77-.768c-.25 0-.47.13-.609.32h-.01a8.51 8.51 0 0 1-6.983 3.646 8.5 8.5 0 0 1-8.502-8.502A8.487 8.487 0 0 1 12 3.41c4.286 0 7.812 3.166 8.402 7.292h-1.03c-.21 0-.329.24-.21.41l1.879 2.627c.1.15.32.14.42 0l1.848-2.627a.26.26 0 0 0-.21-.41" />
+            <path d="m14.52 15.251.01.01.04.04c.11.09.25.14.4.14a.644.644 0 0 0 .48-1.069l-.02-.02-.01-.01-3.047-2.887V6.47a.641.641 0 0 0-1.28 0v5.545l3.417 3.246z" />
+          </svg>
+          <div>
```
colinrotherham: Is this extra `<div>` needed?

docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+            <p class="nhsuk-u-font-weight-bold nhsuk-body-m nhsuk-u-margin-bottom-0">Long stay</p>
+            <p class="nhsuk-body-s nhsuk-u-margin-bottom-0">Plan to stay between 2 to 8 hours</p>
```
colinrotherham: Typically we'd have a `__content` or similar element as the parent here

Then margins can be automatically removed from "top and bottom" child elements

```scss
.app-example__content {
  @include nhsuk-top-and-bottom;
}
```

docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+    {
+      title: 'Orthopaedic appointment',
+      href: '#',
+      titleClasses: 'nhsuk-u-font-size-22',
+      linkAriaLabel: 'Action needed: Orthopaedic appointment on Monday 3 June 2024 at 9:40am, at The Willows, Croydon University Hospital',
+      descriptionHtml: '
+        <p aria-hidden="true" class="nhsuk-u-font-weight-bold nhsuk-u-margin-top-1 nhsuk-u-margin-bottom-3">Monday 3 June 2024 at 9:40am</p>
+        <p aria-hidden="true" class="nhsuk-u-margin-bottom-3">The Willows, Croydon University Hospital</p>
+        <div class="nhsapp-icon-text nhsuk-body-s nhsuk-u-secondary-text-colour">
+          <div class="nhsapp-icon--32 nhsapp-badge-small--as-icon">
```
colinrotherham: Missing the block class to go with these modifiers

```suggestion
          <div class="nhsapp-icon nhsapp-icon--32 nhsapp-badge-small nhsapp-badge-small--as-icon">
```

docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+            <div class="nhsapp-badge-small__indicator" aria-hidden="true"></div>
+          </div>
```
colinrotherham, line 38: Have you got an `nhsappIconText()` macro to write these lines?

docs/examples/cards/card-link-with-footer.njk
```diff
@@ -6,30 +6,44 @@ vueLink: https://nhsappvuecomponentlibraryv1.nonlive.nhsapp.service.nhs.uk/?path
...
+          <span class="nhsuk-u-visually-hidden">New</span>
+          <p class="nhsuk-body-s">Questionnaire and document attached</p>
+        </div>
```
colinrotherham, line 41: These lines are unavailable to screen readers

They're missing from `linkAriaLabel` and sound important enough to disclose?

src/components/card/_card.scss
```diff
@@ -177,6 +177,36 @@ $card-border-radius: 8px;
         border-bottom: $card-border-width solid $card-border-colour;
       }
     }
+
+    // When the last card is a colour variant, keep the rounded border at the bottom
+    // of the second-last card, and place the coloured card slightly behind it.
+    &:has(
+      .nhsapp-card:last-of-type[class*="nhsapp-card--pale-"],
+      .nhsapp-card:last-of-type[class*="nhsapp-card--dark-"]
```
colinrotherham: Worth flagging that [`:last-of-type` means "the last element of its type (tag name)"](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:last-of-type)

So if the second-last card is an `<a>` but the actual last card is a `<div>` it won't work

Probably safer to style `:not(:last-child)` instead?

src/components/card/_card.scss
```diff
@@ -177,6 +177,36 @@ $card-border-radius: 8px;
...
+        .nhsapp-card__container {
+          border-bottom: 0;
+        }
+      }
+
+      .nhsapp-card:last-of-type {
+        border-bottom: 0;
+        border-radius: 0 0 $card-border-radius $card-border-radius;
+        margin-top: -$card-border-radius;
+        padding-top: calc($card-border-radius - $card-border-width);
```
colinrotherham: Does this need `calc()` if they're both pixel units?

```suggestion
        padding-top: $card-border-radius - $card-border-width;
```

src/styles/icon/_icons.scss
```diff
@@ -15,7 +16,12 @@
   fill: nhsuk-colour("black");
 }
 
+.nhsapp-icon--text-colour {
+  fill: currentColor;
+}
+
 .nhsapp-icon--32 {
+  flex: none;
```
colinrotherham: It's a bit of a surprise that only icon `--32` gets `flex: none;`

src/styles/icon/_icons.scss
```diff
@@ -15,7 +16,12 @@
   fill: nhsuk-colour("black");
 }
 
+.nhsapp-icon--text-colour {
+  fill: currentColor;
```
colinrotherham, line 20: You'll probably get the [Safari issue where `:visited` links ignore `currentColor`](https://bugs.webkit.org/show_bug.cgi?id=244025)

docs/examples/cards/card-link-with-footer.njk
imnatgreen: I'd see it more as a description to the heading - I think being a heading might suggest that the location field underneath is a descendant of it, whereas they are closer to siblings in terms of hierarchy (though we're saying date/time is more important)

docs/examples/cards/card-link-with-footer.njk
imnatgreen: turns out we do, it just wasn't used in here before!

docs/examples/cards/card-link-with-footer.njk
imnatgreen: yeah - the icon and this div are in the flex layout that nhsapp-icon-text is providing, we want these paragraphs to follow the standard block flow.

docs/examples/cards/card-link-with-footer.njk
imnatgreen: tbh looks it is getting this from `nhsapp-card__content :last-child` anyway, so this `margin-bottom-0` can be removed!

docs/examples/cards/card-link-with-footer.njk
imnatgreen: this bottom bit actually isn't `aria-hidden` - in the real app we are labelling it slightly differently (and we still have "Document**(s)** and questionnaire**(s)** attached" for now) but we have the same behaviour

src/components/card/_card.scss
imnatgreen: nope, you're right!

src/styles/icon/_icons.scss
imnatgreen: yeah good spot, this is left from before I added `nhsapp-icon-text` - it makes more sense to be a modifier of `.nhsapp-icon` when that parent class is applied

src/styles/icon/_icons.scss
imnatgreen: i think i'll leave this out of scope for now as we don't expect this to be used in a link, and a lot of our links have no visited state anyway

docs/examples/cards/card-link-with-footer.njk
imnatgreen: maybe? for a screen reader user, it could start to become excessive in a long list of cards having multiple heading levels within each card though

docs/examples/cards/card-link-with-footer.njk
colinrotherham: No worries, wasn't sure. Thanks

Can there be a PR review app? Happy to try it locally otherwise

docs/examples/cards/card-link-with-footer.njk
colinrotherham: Just with it being bold it adds a little emphasis (visually) that screen readers won't get

docs/examples/cards/card-link-with-footer.njk
imnatgreen: probably not worth it here, its just `<div class="nhsapp-icon-text [...]">` and is only being used in these 2 places

src/components/card/_card.scss
imnatgreen: yeah very good point - need to investigate this one a little more as it seems we have a related bug on the NHS App homepage where the `border: 0` in `nhsapp-card--with-media` is actually being applied wrong due to the last-of-type div being the feedback banner below it 😅 

Screenshot 2026-08-14 at 17 12 03
Screenshot 2026-08-14 at 17 12 16

imnatgreen: fixed using first/last child across the card now!

docs/examples/cards/card-link-with-footer.njk
imnatgreen: component library is getting deployed to a preview URL (see netlify comment), but not really for the main app as we have to bring the new `nhsapp-frontend` into `nhsapp-vue-component-library`, and upgrade _that_ in the Vue.js app to see changes.

edwardscull (approved):

PR #554, opened 2026-08-18
Add account filled icon

frankieroberto:
The profile icon and link has moved into the bottom navigation / tab bar, and now has a filled variant when active.

Whilst this is a native component, NHS App frontend includes the bottom navigation icons mainly for uses within web-based [NHS App prototypes](https://github.com/NHSDigital/nhsapp-prototype), which mock up the native component in HTML.

Filled variant lifted from https://github.com/nhsuk/nhsapp-frontend/blob/main/src/assets/icons/account.svg (thanks @davidhunter08)

Documentation also updated to mention that the heart and plus icons are no longer used in the bottom navigation, but were previously.


davidhunter08 (approved):
nice one @frankieroberto

PR #561, opened 2026-08-20
Add titleHtml option for cards

### LINKED ISSUE #560: Add title html option for Card component
The card component has an option to set `description` or `descriptionHtml`, but for the title it only has `title`.

This means that to set HTML content for the `title`, which is needed to added the small badge for the unread message indicator, you need to use the `| safe` filter to avoid the HTML being escaped.

Adding an HTML option for title would remove the need for this, and improve consistency of the component.

This will make it easier to do #556

frankieroberto:
This allows you to set HTML within the title section of a card, for example to add an unread badge, without having to use the `| safe` filter.

Resolves #560


davidhunter08 (approved):
LGTM

PR #569, opened 2026-08-25
Add iOS alert component guidance

Graham-Pembrey:
Still to do:

- ~~add examples~~
- ~~add any technical specifics not covered in the design guidance so far~~
- ~~update the 'research' section when we have the findings from the latest research round~~

localhost_8080_ios_alert_


frankieroberto:
@Graham-Pembrey I've updated this a bit so that it now says NOT to use alerts to confirm actions, and instead to use the `confirmationTitle` option on the toolbar buttons (which display the action sheet).

That leaves the alert component a bit thin on examples though - the only one kept in at the moment is the auto-log-out one. Can we think of any others, where an alert might be the right component instead of the confirmation dialog/action sheet?

Graham-Pembrey:
Your changes looked great @frankieroberto. I made some tweaks just now (thanks for fixing the linting!). 

I've been looking into other examples of how we might use alerts. Some possible candidates might be:

- 'Update the NHS App' when users need to update their version (currently we use a native shutter screen for this instead)
- 'Not enough storage space' when users try to download a document, e.g. a GP surgery letter (again, we currently use a shutter screen)

I wonder if for now we should only give the time-out warning example, and aim to build on it? 

I'll also have a think about drafting some action sheet guidance this page could link to, unless anyone has started this already.

frankieroberto:
@Graham-Pembrey I've included the confirmation dialog guidance in this pull request - on the "Toolbar" component page - as that’s where it lives code-wide if it's triggered from the toolbar items. Welcome your feedback on that! (Feel free to make any edits)

frankieroberto (approved):

frankieroberto:
@Graham-Pembrey @davidhunter08 happy if we merge this, or is there anything left to add?

Graham-Pembrey:
@frankieroberto sounds good to me! I think we can keep looking out for more use cases for alerts and add them in over time. I'll check the toolbar guidance today.

PR #582, opened 2026-09-04
App version of expander component

### LINKED ISSUE #581: [Feature request]: An expander component with rounded corners
### Description

The XDX team intend to use the NHS UK expander component to display NHS UK content on the Test Results Details page. The expander looks inconsistent when placed alongside another clickable white card components because there are noticeable differences in borders, corner radius, padding and typography

It would be beneficial to have an expander component with styling consistent with existing NHS App components.

This would support the XDX and Health Information teams avoid placing visually similar interactive components with inconsistent styling side by side.

### Related component

The expander component:
https://service-manual.nhs.uk/design-system/components/expander
https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/29 

### Context

_No response_

### Alternatives

_No response_

### Additional information (if applicable)

_No response_

bharris-nhsd:
Addition of `nhsapp-expander` modifier class and updates to expander component to align with nhs app front end. 

### Changes

Changes have been made to align more closely with the card links:

- Reduced padding
- Added border radius
- Increased border width
- Matched link styling, underline removed from default, underline added to hover

Have not matched the font weight / font size to card links but provided an examples below for review.

[Components/expander page preview](https://deploy-preview-582--nhsapp-frontend.netlify.app/components/expander/)

[Component example preview](https://deploy-preview-582--nhsapp-frontend.netlify.app/examples/details/expander/)

#### Now

Group 4
Group 5

### Before

haemoglobin-14-3-25-xdx-test-6-nhs-expanderhaemoglobin-14-3-25-xdx-test-6-nhs-expander-open

### With bold text

Group 6

### With bold text and same font size

Group 7


src/components/details/_index.scss
```diff
@@ -0,0 +1,53 @@
...
+    border-radius: 0 0 $details-border-radius $details-border-radius;
+  }
+
+  .nhsuk-details__summary-text {
+    text-decoration: none;
+
+    &:hover,
+    &:focus {
+      text-decoration: underline;
+    }
```
frankieroberto, line 33: At the moment the underline appears on hover only when you hover on the text in the summary, not the whitespace surrounding it in the summary (like in the card link).

I think the fix is to nest the CSS something like this:

```
  &:hover,
  &:focus {
    .nhsuk-details__summary-text {
      text-decoration: underline;
    }
  }
```

frankieroberto (changes requested):
Looking good!

One comment on when the underline appears.

Another possible thing to look at: at the moment when you hover on the summary, the border colour for the whole component becomes a darker grey. This is inherited from the nhsuk-component, and does offer an extra sense of 'this is clickable', but it's a different from the card links, which don't do that (probably as they can contain multiple links? and mostly used on touch screens?) - could either turn off the border hover colour change, or leave it as-is.

src/components/details/_index.scss
bharris-nhsd: Thanks for taking a look Frankie. I think after our chat in slack too I am going to put the link style back anyway and keep it distinct from the cards, so it doesn't look like a navigational element. I will revert back to the default, so I won't need to fix the hover issue for the underline. 

I am going to remove the border hover colour change though. Thanks for flagging, I left as is because I assumed the menu would be doing the same on desktop but make sense it wouldn't from what you've said. 

Screenshot 2026-09-14 113521

I'll update when I've make the changes.

bharris-nhsd: @frankieroberto I've made those changes now: [Component example preview](https://deploy-preview-582--nhsapp-frontend.netlify.app/examples/details/expander/)

michaelgallagher:
this looks 95% there to me

one small niggle: the text block inside the expander, when it is open, doesn't quite align with the icon 

(looks like it has about 4–6px extra at the left and right?)

frankieroberto:
> this looks 95% there to me
> 
> one small niggle: the text block inside the expander, when it is open, doesn't quite align with the icon
> 
> (looks like it has about 4–6px extra at the left and right?)

Good spot, thanks. Fixed now.

Have also increased the margin below the expander, from 4 on the responsive scale to 6 - to match the default spacing below card links (or card groups).

frankieroberto (approved):
Looking good!

PR #584, opened 2026-09-09
Experimental: update design system website to use 2-tier top nav

frankieroberto:
This tries out an IA for combining the web, iOS and Android guidance into a single website, using a second-level horizontal navigation.

This would mean that Web has its own sub-nav:

Screenshot 2026-09-09 at 16 35 00

And iOS and Android have their own sub-nav:

Screenshot 2026-09-09 at 16 34 41

...however the 'Community' section and the home page sit across all 3 platforms.

We could also later decide to move 'Patterns' from being a web-specific sub-section to being a top-level section that sits across all platforms, if it proves to be useful to describe patterns in a more abstract way, with content on the pages describing any differences per-platform.

See https://github.com/NHSDigital/native-nhsapp-ucd-team/issues/91

### DIFF
--- a/docs/_includes/layouts/android.njk
+++ b/docs/_includes/layouts/android.njk
@@ -1,0 +1,23 @@
+---
+section: android
+---
+
+{% extends "./sidebar.njk" %}
+{% from 'nhsapp/components/tag/macro.njk' import nhsappTag %}
+
+{% block content %}
+  {{ nhsappTag({
+  text: 'Work in progress',
+  classes: 'nhsapp-tag--blue nhsuk-u-margin-bottom-4'
+  })}}
+
+  {% if tags and "android" in tags %}
+    <span class="nhsuk-caption-xl">Android</span>
+  {% endif %}
+  <h1 class="nhsuk-heading-xl nhsuk-u-margin-bottom-5">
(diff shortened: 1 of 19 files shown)


frankieroberto:
@davidhunter08 @Tosin-Balogun @Graham-Pembrey @michaelgallagher this is not ready to merge at all yet (it's using a not-yet-released component for starters), but let me know what you think of the overall approach.

Tosin-Balogun:
@frankieroberto Nice one! It's an interesting idea. So, if I'm getting it right, rather than now where we use those platform tabs within each component, this time, we create a whole section for each platform?

frankieroberto:
@Tosin-Balogun yep, that’s the idea. It’s sort-of the IA we already have, with the /ios/ section being separate. My thinking is that the 3 platforms are quite divergent, and not all platforms are going to have the same components with the same names, so it makes sense to keep them separate. Although I think we can link between them using links in a right hand column where there are similar components across the different platforms - see crude diagram at https://github.com/NHSDigital/native-nhsapp-ucd-team/issues/91

“Patterns” could be a top-level cross-platform section, or could exist at the platform level — I’m not sure on that one either way yet.

PR #585, opened 2026-09-11
Update homepage to use Hero macro

frankieroberto:
The component macro now supports content in columns (since 10.6), so can be used instead of the raw HTML.

### DIFF
--- a/docs/assets/css/components/_hero.scss
+++ b/docs/assets/css/components/_hero.scss
@@ -5,21 +5,18 @@
 // Hero overrides
 // Reduces spacing on mobile screens
 .app-hero {
-  background-color: nhsuk-colour("blue");
-
-  .nhsuk-hero__wrapper {
+  .nhsuk-hero__content {
     @include nhsuk-media-query($until: tablet) {
       padding-bottom: nhsuk-spacing(5);
       padding-top: nhsuk-spacing(5);
     }
   }
+
+  .nhsuk-image {
+    @include nhsuk-media-query($until: desktop) {
+      display: none;
+    }
+  }  
 }
 
-.app-hero-img {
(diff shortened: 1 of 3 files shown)


docs/assets/css/components/_hero.scss
```diff
@@ -5,21 +5,18 @@
...
       padding-bottom: nhsuk-spacing(5);
       padding-top: nhsuk-spacing(5);
     }
   }
+
+  .nhsuk-image {
+    @include nhsuk-media-query($until: desktop) {
+      display: none;
+    }
+  }
```
colinrotherham, line 19: Want to swap this with `nhsuk-u-display-none-until-desktop`?

We forgot to document the `-until` classes, [only the `-from-` ones](https://service-manual.nhs.uk/design-system/styles/override-classes) 🤦‍♂️

frankieroberto: Nice, done in cd1934e778b5df632380fe86e8758b956e027d9e - this means we no longer need any custom CSS for the hero. 🙌

docs/index.njk
```diff
@@ -2,27 +2,43 @@
...
+    content: [
+      {
+        heading: "Design and build accessible services for the NHS App",
+        width: "one-half",
+        html: heroHtml
+      },
+      {
+        image: {
+          src: "assets/images/hero-image.svg",
+          alt: "",
```
colinrotherham, line 33: Can drop the default `alt: ""` if you like

```suggestion
```

colinrotherham (approved):

docs/index.njk
frankieroberto: I quite like keeping it to show that it’s intentionally blank rather than just forgotten… 😃

PR #586, opened 2026-09-14
iOS title component guidance

davidhunter08:
## Description

Adding guidance for the iOS title component.

Related PR: https://github.com/NHSDigital/nhsapp-design-system-ios/pull/107

[Guidance preview](https://deploy-preview-586--nhsapp-frontend.netlify.app/ios/title/)

## Notes

The guidance mentions section header in the when not to use section. I we should write guidance for section headers and then link the guidance?

### DIFF
--- a/docs/ios/title.md
+++ b/docs/ios/title.md
@@ -1,0 +1,149 @@
+---
+layout: layouts/ios.njk
+title: Title
+tags:
+  - iosComponents
+---
+
+The title is the main heading of a screen. It can also appear in the navigation bar.
+
+## When to use
+
+Use a title on every screen, so users know where they are.
+
+Most NHS App screens lead with a large heading at the top of the content, which moves into the navigation bar as the user scrolls down.
+
+## When not to use
+
+Do not use a title for headings within a screen — use a [section header](/ios/section-headers-and-footers).
+
+Do not use more than one title on a screen.
+
+## How it works
+
+A title is made of 2 parts that work together:
+
+- the `Title` component, which draws the large heading in the screen's content
+- the `nhsTitle()` modifier, which sets the heading text and puts it in the navigation bar
+
+There are 2 display modes:
+
+- **large** — the heading is shown at the top of the content, and fades into the navigation bar once the user scrolls past it. This is the default, and how most NHS App screens work.
+- **inline** — the heading is shown in the navigation bar only. Use it on screens that do not lead with a large heading, such as a pushed detail screen.
+
+A title can also have a subtitle and a decorative icon.
+
+The layout adapts to the user's settings:
+
+- text scales with Dynamic Type
+- at large text sizes, the icon is no longer shown so the heading has the full width
+- headings are never truncated: they wrap onto more lines
+
+## How to use
+
+Apply `nhsTitle()` to the screen's scrolling container, and lead the content with a `Title`:
+
+{% from "details/macro.njk" import details %}
+{% call details({ summaryText: "Swift options" }) %}
+{% include "ios/title/swift-options.md" %}
+{% endcall %}
+
+```swift { .nhsuk-code--button }
+ScrollView {
+    Title()
+
+    // the rest of the screen
+}
+.nhsTitle("Health choices")
+```
+
+`Title()` takes its heading from `nhsTitle()`, so the text is written once.
+
+### Titles in the navigation bar only
+
+On a screen that does not lead with a large heading, use `.inline`. A `Title()` is not needed in the content:
+
+```swift { .nhsuk-code--button }
+List {
+    // the rest of the screen
+}
+.nhsTitle("Removed messages", displayMode: .inline)
+```
+
+### Subtitles and icons
+
+Add a subtitle for supporting information about the whole screen.
+
+```swift { .nhsuk-code--button }
+ScrollView {
+  Title(
+    subtitle: "Manage another person's healthcare or give access to your profile to someone you trust."
+  )
+
+  // the rest of the screen
+}
+.nhsTitle("Family and carer access")
+```
+
+### Icons
+
+An icon where it helps users recognise the section [something about only using it for hub screens.]
+
+```swift { .nhsuk-code--button }
+ScrollView {
+  Title(
+      systemImage: "pills.fill"
+  )
+
+  // the rest of the screen
+}
+.nhsTitle("Prescriptions")
+```
+
+### Titles in a list view
+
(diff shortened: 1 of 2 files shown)


docs/ios/title.md
```diff
@@ -0,0 +1,146 @@
+---
+layout: layouts/ios.njk
+title: Title
+tags:
+  - iosComponents
+---
+
+The title is the main heading of a screen. It can also appear in the navigation bar.
```
Graham-Pembrey: I noticed the[ toolbar guidance from Apple](https://developer.apple.com/design/human-interface-guidelines/toolbars) mentions that the toolbar can include 'The title of the current view'. Should we be saying 'It can also appear in the toolbar'? Or am I mixing things up? Think I'm probably mixing things up :-D 

If 'navigation bar' is correct, I wonder if some readers may be unsure what part of the UI it refers to, and whether we should briefly explain that. For example: 'It can also appear in the navigation bar, which shows at the top of the screen when users scroll.'

docs/ios/title.md
```diff
@@ -0,0 +1,146 @@
...
+
+## When to use
+
+Use a title on every screen, so users know where they are.
+
+## When not to use
+
+Do not use a title for headings within a screen, use a section header instead.
```
Graham-Pembrey:
```suggestion
Do not use a title for headings within a screen. Use a section header instead.
```

docs/ios/title.md
```diff
@@ -0,0 +1,146 @@
...
+There are 2 display modes:
+
+- [large title](#large-title)
+- [inline title](#inline-title)
+
+## How to use
+
+### Large title
+
+Large titles display at the top of the content, which moves into the navigation bar as the user scrolls down.
```
Graham-Pembrey:
```suggestion
Large titles display at the top of the content, then move into the navigation bar as the user scrolls down.
```

docs/ios/title.md
```diff
@@ -0,0 +1,146 @@
...
+  )
+
+  // the rest of the screen
+}
+.nhsTitle("Family and carer access")
+```
+
+### With an icon
+
+An icon where it helps users recognise a section. Only use these on hub screens.
```
Graham-Pembrey:
```suggestion
An icon may help users recognise a section. Only use these on hub screens.
```

docs/ios/title.md
```diff
@@ -0,0 +1,146 @@
...
+List {
+    Title()
+        .listRowInsets(EdgeInsets())
+        .listRowBackground(Color.clear)
+        .listRowSeparator(.hidden)
+
+    // the rest of the screen
+}
+.nhsTitle("Face ID")
+```
```
frankieroberto: Should these 3 variants come under the 'Large title' section as level 4 headings, as they don't apply to the 'Inline title'?  It a bit confusing at the mo as the links at the to mention 2 options, but then there are 4 h3 headings?

Graham-Pembrey (changes requested):
This is looking great. I've left a few suggestions and a question.

docs/ios/title.md
frankieroberto: We've not really described 'hub' screens elsewhere (for iOS), and I'm not sure if we’ll keep that term or not?

Could make it more generic and say something like only include it on the first screen navigated to from the home menu, and not any subsequent screens?

docs/ios/title.md
davidhunter08: I think 'toolbar' is the correct term @Graham-Pembrey and so we should replace 'navigation bar' with it.

davidhunter08: Wondering if/how we should mention the inline title in the [toolbar guidance](https://design-system.nhsapp.service.nhs.uk/ios/toolbar/)?

docs/ios/title.md
frankieroberto: Worth mentioning that this emulates the native style and behaviour for iOS for large titles? Although I can't find a great link for it in their guidelines - this seems to be the most relevant? https://developer.apple.com/design/human-interface-guidelines/toolbars#Titles

docs/ios/title.md
```diff
@@ -0,0 +1,146 @@
...
+- [large title](#large-title)
+- [inline title](#inline-title)
+
+## How to use
+
+### Large title
+
+Large titles display at the top of the content, which moves into the navigation bar as the user scrolls down.
+
+<img src="/assets/images/ios/title-large.png">
```
frankieroberto, line 33: Use the markdown syntax for images? And add some alt text.

```suggestion
![Screenshot of...](/assets/images/ios/title-large.png)
```

docs/ios/title.md
davidhunter08: I've changed it to:

H3 - Large title
H4 - With subtitle
H4 - With Icon
H4 - In a list view
H3 - Inline title

So it still says 2 display modes (large, inline).

docs/ios/title.md
davidhunter08: "Only use it on the first screen the user opens from the home menu, not on later screens." ?

docs/ios/title.md
davidhunter08: It looks like all the other iOS guidance pages use HTML, eg: ``, so maybe it's something we want to do on a separate PR, so everything is consistent?

docs/ios/title.md
davidhunter08: I'm not sure tbh, I don't think that guidance tells you much. Do we just leave out the link, mention that it emulates the behaviour but also enhances it by not truncating the large title?

docs/ios/title.md
frankieroberto: At the moment you need to do HTML if you specify the width as 320, but you can use the markdown where it's full-width.

I'm not sure how best to do the 320 width ones within markdown yet - could either add some sort of custom extension, or keep it full width but add some whitespace to either side of the image itself.

docs/ios/title.md
```diff
@@ -0,0 +1,146 @@
+---
+layout: layouts/ios.njk
+title: Title
+tags:
+  - iosComponents
+---
+
+The title is the main heading of a screen. It can also appear in the toolbar.
```
Graham-Pembrey:
```suggestion
The title is the main heading of a screen. It can also appear in the [toolbar](/ios/toolbar/).
```

Graham-Pembrey (approved):
LGTM. I've suggested a cross-link to the toolbar guidance.

PR #588, opened 2026-09-15
Standardise component Sass partials on _index.scss

frankieroberto:
Rename the badge, card, tag and timeline component Sass files to `_index.scss`, matching the button and summary-list components. `all.scss` now forwards each component by directory rather than by an explicit filename.

The old file names (e.g. components/badge/badge) are kept as deprecated stub files that forward to `_index.scss`, so existing `@import`/`@use` paths keep working. They can be removed in a future breaking change release.

Also updates the npm install docs to reference the new component paths (this is already partially out of date).


frankieroberto:
@colinrotherham does this seem reasonable to do as a non-breaking change?

colinrotherham (changes requested):
I think there's a bit of a trick to this otherwise we lose the file history 😔

Can you rename (for example) **_badge.scss** → **_index.scss** in one branch

Then in another branch add the old files back in as stubs

Merge the two together

Hopefully it doesn't involve (temporarily) using different file names to avoid conflicts, but give it a go 🙏

End of commit history

PR #589, opened 2026-09-15
Update toolbar guidance

Graham-Pembrey:
- Added a 'Title' section (is it right to list this as a toolbar item like the others?)
- Updated the confirmation dialog prescriptions example text
- Changed 'modal' to web overlay in a couple of places and linked to the relevant guidance on isolated web journeys (see Slack convo with Mark Robinson)
- Added writing guidance for confirmation dialogs (maybe an example could go here, or above?)
- A couple of other small tweaks

[Preview link](https://deploy-preview-589--nhsapp-frontend.netlify.app/ios/toolbar/)


docs/ios/toolbar.md
```diff
@@ -51,11 +52,19 @@ If you need to add a different type of toolbar action, use one of these generic
...
 The back button will appear in the toolbar at the top left automatically when using the `NavigationStack`. You do not need to add this.
 
+### Title
+
+Titles in the toolbar can help users understand where they are in the app.
+
+If there is a large title on the screen, it should move into the toolbar as the user scrolls down. On a screen that does not lead with a title, use an inline title that stays in the toolbar.
+
+See our [title guidance](/ios/title/) for more details.
+
```
frankieroberto, line 62: Love this type of cross-linking! ❤️

docs/ios/toolbar.md
```diff
@@ -94,7 +103,7 @@ struct BookAppointmentView: View {
 
 <img src="/assets/images/ios/done-button.png" width="320" alt="Screenshot showing a green button on the right of a mobile screen with the word 'Done' inside it in white text">
 
-The 'done' button is added to any screens presented as modal, when the user has reached the end of a journey and has completed a task. For example, after booking an appointment.
+The 'done' button is added to any screens presented as [web overlay](/patterns/access-web-journeys/), when the user has reached the end of a journey and has completed a task. For example, after booking an appointment.
```
frankieroberto: I like the idea of linking to the web overlay here, but can we also mention that it should be used on any native views presented as a sheet (or modal, or whatever we call it)?

docs/ios/toolbar.md
```diff
@@ -318,6 +327,14 @@ struct RemovedMessageView: View {
 }
 ```
 
+## Writing confirmation dialogs
+
+You can use confirmation dialogs to make sure users understand the consequences of an action, and to give them a chance to cancel before they continue.
+
+The text should be a single sentence. It should give any important extra context in a brief and direct way.
+
+Use verbs for button text. If you have more than one button, make sure that the difference between the options is clear.
```
frankieroberto: At the moment we've not given you any choice in what the button text is - it’s always the same label as the toolbar button itself (either the text shown in the text toolbar item, or the accessible label for the icon toolbar item).

We could let you override this if they ought to be different - but don’t think we’ve seen a need for that yet?

Graham-Pembrey: That makes sense! I wouldn't have thought there would be much need to override it.  Maybe we can take the button text part out for now, but come back to it if we find new examples where that could happen.

frankieroberto: We decided to drop this line for now, as you can't currently change the button text or add a second button.

```suggestion
```

frankieroberto (approved):
🚀

PR #590, opened 2026-09-16
Update publish workflow

frankieroberto:
This updates the publish workflow to:

* avoid using the third-party `softprops/action-gh-release` action and instead use the `gh release upload` CLI command (`gh` is pre-installed)
* fix the publishing of pre-release versions


.github/workflows/publish.yaml
```diff
@@ -26,12 +26,17 @@ jobs:
...
+            "./dist/nhsapp-frontend-${{ steps.get_version.outputs.VERSION }}.zip" \
+            --clobber
 
       - name: Publish NPM package
-        run: npm publish --access public --provenance
-        env:
-          IS_PRERELEASE: ${{ github.event.release.prerelease }}
+        run: |
+          if [[ "${{ github.event.release.prerelease }}" == "true" ]]; then
+            npm publish --tag next --provenance --access public
```
colinrotherham: I wondered why we didn't have `--provenance` on NHS.UK frontend

From the npm [**Generating provenance statements:** Prerequisites](https://docs.npmjs.com/generating-provenance-statements#prerequisites) guidance

>**Note:** If you use [trusted publishing](https://docs.npmjs.com/trusted-publishers), provenance attestations are automatically generated for your packages without requiring the `--provenance` flag. This provides enhanced security and eliminates the need for access tokens in your CI/CD workflows.

.github/workflows/publish.yaml
```diff
@@ -26,12 +26,17 @@ jobs:
...
+          gh release upload "${{ github.event.release.tag_name }}" \
+            "./dist/nhsapp-frontend-${{ steps.get_version.outputs.VERSION }}.zip" \
+            --clobber
 
       - name: Publish NPM package
-        run: npm publish --access public --provenance
-        env:
-          IS_PRERELEASE: ${{ github.event.release.prerelease }}
+        run: |
+          if [[ "${{ github.event.release.prerelease }}" == "true" ]]; then
```
colinrotherham: Are we preferring this `if [[` style to NHS.UK frontend and the React components?

```yaml
      - name: Publish npm package
        if: ${{ !github.event.release.prerelease }}
        run: npm publish --tag latest

      - name: Publish npm package (pre-release)
        if: ${{ github.event.release.prerelease }}
        run: npm publish --tag next
```

.github/workflows/publish.yaml
frankieroberto: Ah nice - we can drop this then. 👍

frankieroberto:
```suggestion
            npm publish --tag next --access public
```

.github/workflows/publish.yaml
```diff
@@ -26,12 +26,17 @@ jobs:
...
+            npm publish --tag next --provenance --access public
+          else
+            npm publish --tag latest --provenance --access public
```
frankieroberto:
```suggestion
            npm publish --tag latest --access public
```

.github/workflows/publish.yaml
```diff
@@ -26,12 +26,17 @@ jobs:
...
-        uses: softprops/action-gh-release@v3
-        if: startsWith(github.ref, 'refs/tags/')
-        with:
-          files: ./dist/nhsapp-frontend-${{ steps.get_version.outputs.VERSION }}.zip
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload "${{ github.event.release.tag_name }}" \
+            "./dist/nhsapp-frontend-${{ steps.get_version.outputs.VERSION }}.zip" \
+            --clobber
```
colinrotherham: Shall we use more envs like in [**Using GitHub CLI in workflows**](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli)?

Do we definitely want `--clobber` here?

Was it only added just in case the npm release failed and the action was run again?

I think it'd be safer (perhaps) to manually delete the release in that case

```suggestion
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          TAG: github.event.release.tag_name
          VERSION: steps.get_version.outputs.VERSION

        run: gh release upload "$TAG" "./dist/nhsapp-frontend-$VERSION.zip"
```

.github/workflows/publish.yaml
frankieroberto: Ooh good shout - I prefer the other style, as it avoids the idiosyncratic `if / else / fi` syntax from bash 😄 

Have updated.

colinrotherham: Cheers, and you can see (from the UI) which named step was run

.github/workflows/publish.yaml
frankieroberto: I wasn't sure on this either. I guess in theory it could be useful if the workflow failed for some reason and you need to re-try it.

But does feel a bit unexpected. Will remove.

.github/workflows/publish.yaml
```diff
@@ -26,12 +26,16 @@ jobs:
         run: echo "VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
 
       - name: Release
-        uses: softprops/action-gh-release@v3
-        if: startsWith(github.ref, 'refs/tags/')
```
colinrotherham, line 30: There's some merit in this, but can see why it was necessary

Unless we want to use `if: startsWith(github.ref, 'refs/tags/v')`?

(Note the check for a `v` prefix)

I know over on NHS.UK frontend we're responding to `push.tags` but it meant we could use:

```yaml
on:
  push:
    tags:
      - 'v*'
```

We've got to think about someone accidentally:

1. Making a new release without the `v` prefix
2. Making a new release with a tag that doesn't match **package.json** `version`

frankieroberto: Maybe it's safer to parse the `package.json` and get the version number direct from there? 🤔

frankieroberto: I've added a job-level check on the tag starting with v for now: 71cff64a8b40f2c4719a432461fdc04160c4d705

colinrotherham (approved):

PR #591, opened 2026-09-16
Add iOS Toggle component guidance

davidhunter08:
Guidance for:
- https://github.com/NHSDigital/nhsapp-design-system-ios/pull/120

[Preview link](https://deploy-preview-591--nhsapp-frontend.netlify.app/ios/toggle/)


docs/ios/toggle.md
```diff
@@ -0,0 +1,84 @@
...
+@State private var acceptsCookies = true
+
+var body: some View {
+    Form {
+        Toggle(isOn: $acceptsCookies) {
+            Text("Allow optional analytic cookies")
+                .bold()
+            Text("I accept the use of optional analytic cookies used to improve the performance of the NHS App")
+                .foregroundStyle(.nhsSecondaryText)
+        }
```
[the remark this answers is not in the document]
frankieroberto: Is this copilot gaslighting us? 😄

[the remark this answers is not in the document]
davidhunter08: That's not true.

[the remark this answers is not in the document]
davidhunter08: I think we replied at the same time 😂

Screenshot 2026-09-16 at 14 19 04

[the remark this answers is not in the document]
davidhunter08: Definitely gaslighting us. I took the screenshot from that code in Xcode.

docs/ios/toggle.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+
+Where a toggle has a subtitle, keep the title short and put the detail in the subtitle, so the on or off state is clear at a glance.
+
+## Accessibility
+
+This component supports Dynamic Type, Dark Mode and VoiceOver.
+
+The style keeps the native toggle rather than drawing a custom control, so it inherits the system switch's accessibility, haptics and disabled appearance.
```
frankieroberto, line 78: Worth mentioning that it also supports the accessibility option for "On/Off Labels" (Accessibility -> Display and text Size).

Could even include a screenshot?

docs/ios/toggle.md
```diff
@@ -0,0 +1,84 @@
...
+  - iosComponents
+---
+
+Toggles let users switch a single option on or off.
+
+<img src="/assets/images/ios/toggle.png" width="375">
+
+## How it works
+
+There is one style, applied with the `.nhsToggle` preset. It uses the native iOS toggle, tinted NHS blue when on.
```
frankieroberto, line 14: Link to the Apple docs here?

```suggestion
There is one style, applied with the `.nhsToggle` preset. It uses the native [SwiftUI Toggle](https://developer.apple.com/documentation/swiftui/toggle), tinted NHS blue when on.
```

docs/ios/toggle.md
```diff
@@ -0,0 +1,84 @@
+---
+layout: layouts/ios.njk
+title: Toggles
+tags:
+  - iosComponents
+---
+
+Toggles let users switch a single option on or off.
```
frankieroberto: Could link to https://developer.apple.com/design/human-interface-guidelines/toggles here somehow?

docs/ios/toggle.md
```diff
@@ -0,0 +1,84 @@
...
+
+<img src="/assets/images/ios/toggle.png" width="375">
+
+## How it works
```
frankieroberto, line 12: Should we mention in this section, or somewhere, that with toggles it should apply immediately, with no confirmation or submit button?

ie don't use it within a form of a multi-step transactional flow?  Or are there cases where this is ok?

(We can park this if it's a can of worms)

frankieroberto (approved):
Looking good. Left some suggestions, but no blockers.

docs/ios/toggle.md
```diff
@@ -0,0 +1,84 @@
...
+    }
+    .toggleStyle(.nhsToggle)
+}
+```
+
+### Toggle with a subtitle
+
+<img src="/assets/images/ios/toggle-subtitle.png" width="375">
+
+When a toggle needs supporting text, give it a two-part label.
```
Graham-Pembrey, line 54: Think 'supporting text' here is fine, but another option could be to call it 'hint text'. It's serving the same purpose as the hint text form element in the service manual: giving extra info about what a choice means. < Not necessarily thinking this needs to change, just pondering it!

docs/ios/toggle.md
Graham-Pembrey:
```suggestion
Toggles let users switch a single option on or off. They are a standard iOS component (see [Apple's toggle guidance](https://developer.apple.com/design/human-interface-guidelines/toggles)). 
```

docs/ios/toggle.md
Graham-Pembrey: RE my comment from yesterday here: I think maybe leave this as-is for now.

docs/ios/toggle.md
```diff
@@ -0,0 +1,84 @@
...
+
+Where a toggle has a subtitle, keep the title short and put the detail in the subtitle, so the on or off state is clear at a glance.
+
+## Accessibility
+
+This component supports Dynamic Type, Dark Mode and VoiceOver.
+
+The style keeps the native toggle rather than drawing a custom control, so it inherits the system switch's accessibility, haptics and disabled appearance.
+
+VoiceOver announces the toggle's label and its on or off state, and disabled toggles fade visually but remain in the accessibility tree, so users know the option exists even when it is not available.
```
Graham-Pembrey:
```suggestion
VoiceOver announces the toggle's label and its on or off state. Disabled toggles fade visually but remain in the accessibility tree, so users know the option exists even when it is not available.
```

docs/ios/toggle.md
Graham-Pembrey: I think that might be worth adding. Something along the lines of what you've mentioned:

> Toggles should take effect immediately. Users should not need to confirm their choice using a separate "Continue" or "Done" button.

It could be reconsidered if anyone has evidence or a use case that challenges it, but it seems right and like the general consensus? The related Android [switch ](https://m3.material.io/components/switch/guidelines)component actually specifies that "Switches change settings and other options immediately".

Graham-Pembrey (approved):
LGTM! A few suggestions and comments.

docs/ios/toggle.md
davidhunter08: Simulator Screenshot - iPhone 17 Pro Max - 2026-09-18 at 10 36 45

In the native flow for request a repeat prescription we do use toggles for selecting medicines 🤔

Graham-Pembrey: Oh yeah! Hmmm. And that's definitely tested well, so maybe we should leave out adding the "takes effect immediately" line to the guidance... It's not something that's mentioned in the Apple guidance, from what I can see.

docs/ios/toggle.md
davidhunter08: Graham added a link for this 👍

docs/ios/toggle.md
davidhunter08: Done

docs/ios/toggle.md
davidhunter08: Shall we leave this for now and tackle it when we start making forms?

Graham-Pembrey: Sounds like a good plan!

PR #594, opened 2026-09-18
Add scss stub files

frankieroberto:
This adds backwards-compatibility for the old SCSS partials


src/components/badge/_badge.scss
```diff
@@ -0,0 +1,2 @@
+// Deprecated: import "components/badge" instead. Kept for backwards compatibility.
+@forward "index";
```
colinrotherham: We simply use ~`"index"`~ `"."` on NHS.UK frontend, now that index is the default

```suggestion
@forward ".";
```

frankieroberto: Done

colinrotherham (approved):

